PR #15, opened 2019-04-29
Feature/DispatchTimer

borut-t:


PovioKit/Classes/Utilities/DispatchTimer/DispatchTimer.swift
```diff
@@ -0,0 +1,53 @@
...
+//  PovioKit
+//
+//  Created by Borut Tomažin on 11/12/2018.
+//  Copyright © 2019 Povio Labs. All rights reserved.
+//
+
+import Foundation
+
+/// A NSTimer replacement using GCD.
+class DispatchTimer {
```
tonikocjan: I think this class needs to be explicitly marked `public` as well as it's _public_ methods.

PR #20, opened 2019-05-05
Feature/RestClient

tonikocjan:
`RestClient`


PovioKit/Classes/Networking/RestClient/Host.swift
```diff
@@ -0,0 +1,38 @@
...
+  case production
+}
+
+public extension Host {
+  var baseURL: String {
+    switch self {
+    case .development:
+      return "https://us-central1-waybots-a2558.cloudfunctions.net"
+    case .production:
+      return "https://us-central1-waybots-production.cloudfunctions.net"
```
borut-t: We are using fixed hosts here. This should go away. Also, we should think about making this generic enough so implementation can define `baseUrl` on its own.

PovioKit/Classes/Networking/RestClient/EndpointProtocol.swift
```diff
@@ -0,0 +1,47 @@
...
+  var retryCount: Int {
+    return 0
+  }
+  
+  var apiVersion: Version {
+    return .v1 // Currently not in use
+  }
+  
+  var authentication: Authentication {
+    return .authenticated // Currently not in use
```
borut-t: Please remove the comment.

PovioKit/Classes/Networking/RestClient/EndpointProtocol.swift
```diff
@@ -0,0 +1,47 @@
...
+  var url: String {
+    return "\(Host.current.baseURL)/\(path)"
+  }
+  
+  var retryCount: Int {
+    return 0
+  }
+  
+  var apiVersion: Version {
+    return .v1 // Currently not in use
```
borut-t: Please remove the comment.

PovioKit/Classes/Networking/RestClient/Host.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+//  Created by Borut Tomažin on 14/01/2019.
+//  Copyright © 2019 Povio Labs. All rights reserved.
+//
+
+import Foundation
+
+public enum Host {
+  case development
+  case production
```
borut-t: This should not be hardcoded. We should think about making that more generic.

PovioKit/Classes/Networking/RestClient/Host.swift
```diff
@@ -0,0 +1,38 @@
...
+    switch self {
+    case .development:
+      return "DEV"
+    case .production:
+      return "PROD"
+    }
+  }
+  
+  static var current: Host {
+    return Constants.Environment.isDevTarget ? .development : .production
```
borut-t: This whole class should be generic. I'm thinking in a way to pass info to the rest client setup somehow on app launch 🤔

PovioKit/Classes/Networking/RestClient/RestClient.swift
```diff
@@ -0,0 +1,160 @@
...
+            case false:
+              strongSelf.retryCounter.clear(for: endpoint)
+              result?(.failure(.generic(error)))
+            }
+          }
+      }
+    }
+  }
+  
+  func updateToken(_ completion: ((_ token: String?) -> Void)?) {
```
borut-t: This method is firebase specific. We should throw it away for now.
And, we should think more about that to maybe define it at config stage 🤔

PovioKit/Classes/Networking/RestClient/RestClient.swift
```diff
@@ -0,0 +1,160 @@
...
+//
+//  Created by Borut Tomažin on 14/01/2019.
+//  Copyright © 2019 Povio Labs. All rights reserved.
+//
+
+import Foundation
+import Alamofire
+
+public class RestClient {
+  private let manager: Alamofire.SessionManager
```
borut-t: Can we try the solution without `Alamofire` only using `UrlHttpSession` instead?

PovioKit/Classes/Networking/RestClient/RestClient.swift
```diff
@@ -0,0 +1,160 @@
...
+    if let token = token {
+      headers["Authorization"] = "Bearer \(token)"
+    } else {
+      headers["Authorization"] = nil
+    }
+    
+    // app specific headers
+    headers["X-SKIP-APP-PLATFORM"] = "iOS"
+    headers["X-SKIP-APP-NAME"] = "renter"
+    headers["X-SKIP-APP-VERSION-STRING"] = UIDevice.appVersion
```
borut-t: Remove app specific headers. We should also define generic headers at the setup stage.

PovioKit/Classes/Networking/RestClient/RestClientProtocol.swift
```diff
@@ -0,0 +1,20 @@
...
+
+public protocol RestClientProtocol {
+  typealias Headers = [String: String]
+  typealias Params = [String: Any]
+  typealias DataResult = ((Result<DataResponse, NetworkError>) -> Void)?
+  
+  func GET(endpoint: EndpointProtocol, parameters: Params?, headers: Headers?, _ result: DataResult)
+  func GET<T: Decodable>(decode: T.Type, endpoint: EndpointProtocol, parameters: Params?, headers: Headers?, _ result: ((Result<T, NetworkError>) -> Void)?)
+  func POST(endpoint: EndpointProtocol, parameters: Params?, headers: Headers?, _ result: DataResult)
+  func POST<T: Decodable>(decode: T.Type, endpoint: EndpointProtocol, parameters: Params?, headers: Headers?, _ result: ((Result<T, NetworkError>) -> Void)?)
```
borut-t: We'll probably need to support other methods, like `PUT` and `DELETE`.

PovioKit.podspec
```diff
@@ -62,4 +62,13 @@ Pod::Spec.new do |s|
       cs.source_files = 'PovioKit/Classes/Views/GradientView/**/*'
     end
   end
+  
+  s.subspec 'Networking' do |sp|
+    sp.source_files = 'PovioKit/Classes/Networking/**/*'
+    
+    sp.subspec 'RestClient' do |cs|
+      cs.source_files = 'PovioKit/Classes/Networking/RestClient/**/*'
+      cs.dependency 'Alamofire'
```
borut-t: I don't like this dependency in our pods. What do you think about switching to url session instead?

tonikocjan: Why not? We've been using Alamofire in most of our projects thus far...

tonikocjan: What I propose instead is that we make `RestClient` as general as possible, meaning that we also abstract away the underlying HTTP request engine. 
We could then implement two (or more) such interfaces: `URLSession` implementation and `Alamofire` implementation. Users of the lib could then decide which implementation they want: `pod 'PovioKit/Networking/RestClient/Alamofire' or 
`pod 'PovioKit/Networking/RestClient/URLSession'
What do you think?

borut-t: Yes, Alamofire should be opt-in. Let's proceed.

PR #40, opened 2019-11-28
Feature/Alamofire Network Client

tonikocjan:
A new approach towards integrating Alamofire as a core API client.
[Examples.](https://github.com/poviolabs/PovioKit/blob/feature/alamofire-network-client/PovioKit/Classes/Networking/AlamofireNetworkClient/AlamofireRestClient.md)


PovioKit.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'PovioKit'
-  s.version          = '0.3.0'
+  s.version          = '0.3.1'
```
borut-t: We'll bump the spec version when we prepare a release branch. Please revert it.

tonikocjan: Reverted.

PR #45, opened 2019-12-06
Feature/Add tests for Throttler

klemenzagar91:


Example/Tests/Utilities/Throttler/ThrottlerTests.swift
```diff
@@ -0,0 +1,84 @@
...
+  func testShouldExecuteWhenDelayed() {
+    let delay = 100
+    let waiting = delay + 10
+    let sut = Throttler(delay: .milliseconds(delay))
+    let expectation = self.expectation(description: "Executed")
+    var output: [String] = []
+    sut.execute {
+      output.append("A")
+    }
+    DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(waiting)) {
```
borut-t: Using dispatch for the wait is not a good practice. We should mock Throttler instead.

klemenzagar91: Great. I'll improve it.

PR #46, opened 2019-12-06
Feature/Broadcast tests

klemenzagar91:


Example/Tests/Utilities/Broadcast/BroadcastTests.swift
```diff
@@ -0,0 +1,101 @@
...
+  func run()
+}
+
+class MockedListener: MockedProtocol {
+  var executingCount = 0
+  
+  func run() {
+    executingCount += 1
+  }
+}
```
borut-t: Make this private.

PR #47, opened 2019-12-06
Feature/Optional Tests

klemenzagar91:


Example/Tests/Extensions/Foundation/OptionalTests.swift
```diff
@@ -0,0 +1,42 @@
...
+//  Copyright © 2019 CocoaPods. All rights reserved.
+//
+
+import XCTest
+
+class OptionalTests: XCTestCase {
+  
+  func testArrayNilOrEmptyTrueWhenNil() {
+    let sut: [String]? = nil
+    XCTAssert(sut.isNilOrEmpty, "Should not return false when collection is nil")
```
borut-t: I would rather test against boolean value by default, e.g `XCTAssertTrue(sut.isNilOrEmpty, ...)`.

PR #57, opened 2020-01-31
Feature/PromiseKit API Improvements

tonikocjan:
`Promise` now accepts only one generic type argument, which simplifies usage as we don't have to deal with different error types. Also added support for the user to decide on which `DispatchQueue` the promise should be resolved.

### DIFF
--- a/PovioKit/Classes/Utilities/PromiseKit/Promise.swift
+++ b/PovioKit/Classes/Utilities/PromiseKit/Promise.swift
@@ -199,55 +199,6 @@ public extension Promise {
       }
     }
   }
-  
-  /// Returns a new Promise, combining multiple `Promise`s.
-  ///
-  /// Use this method when you need to combine the result of several promises of some type `T`.
-  ///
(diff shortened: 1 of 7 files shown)


PovioKit/Classes/Utilities/PromiseKit/Promise.swift
```diff
@@ -186,68 +162,43 @@ public extension Promise {
   ///   instance.
   /// - Returns: A `Promise` with the result of evaluating `transform`
   ///   as the new success value if this instance represents a success.
-  func map<U>(with transform: @escaping (Value) -> U) -> Promise<U, Error> {
-    chain { Promise<U, Error>(fulfill: transform($0)) }
+  func map<U>(
+    on dispatchQueue: DispatchQueue = .main,
+    with transform: @escaping (Value) throws -> U) -> Promise<U>
```
borut-t:
```swift
func map<U>(on dispatchQueue: DispatchQueue = .main,
            with transform: @escaping (Value) throws -> U) -> Promise<U>
```

PR #60, opened 2020-02-18
Chore/Use Alamofire 5

tonikocjan:


PovioKit.podspec
```diff
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.source_files = 'PovioKit/Classes/**/*.swift'
   s.frameworks = 'UIKit', 'Foundation'
   s.default_subspecs = 'Utilities', 'Extensions', 'Views'
-  s.dependency 'Alamofire', '5.0.0-rc.3'
+  s.dependency 'Alamofire', '5.0.0'
```
borut-t: We should add a dependency to the subspec instead. Can we try that?

tonikocjan: Done.

PR #61, opened 2020-02-24
Feature/URL extension

borut-t:
Extension to simplify building URL with parameters.

### DIFF
--- a/PovioKit/Classes/Extensions/Foundation/URL+Povio.swift
+++ b/PovioKit/Classes/Extensions/Foundation/URL+Povio.swift
@@ -13,6 +13,17 @@ public extension URL {
     guard let string = string else { return nil }
     self.init(string: string)
   }
+  
+  /// Append parameter to the URL
+  func appending(_ name: String, value: String?) -> URL {
+    guard var components = URLComponents(url: self, resolvingAgainstBaseURL: true) else { return absoluteURL }
+    var queryItems = components.queryItems ?? []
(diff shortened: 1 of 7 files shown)


PovioKit/Classes/Extensions/Foundation/URL+Povio.swift
```diff
@@ -13,6 +13,17 @@ public extension URL {
...
   }
+  
+  /// Append parameter to the URL
+  func appending(_ name: String, value: String?) -> URL {
+    guard var components = URLComponents(url: self, resolvingAgainstBaseURL: true) else { return absoluteURL }
+    var queryItems = components.queryItems ?? []
+    let newQueryItem = URLQueryItem(name: name, value: value)
+    queryItems.append(newQueryItem)
+    components.queryItems = queryItems
+    components.percentEncodedQuery = components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
```
nejcvivod: Do we have any other symbols that would need to be encoded? 🤔

borut-t: Great question. I think we should have. Let me think about that.

borut-t: @nejcvivod it looks like components do its job fairly good. If I recall correctly I added percent handling for `+` because it isn't escaped by default 🤷‍♂

PR #62, opened 2020-02-24
Feature/Result success void extension

borut-t:
Simplify calling success with void value.

### DIFF
--- a/Result+Povio.swift
+++ b/Result+Povio.swift
@@ -1,0 +1,15 @@
+//
+//  Result+Povio.swift
+//  Alamofire
+//
+//  Created by Borut Tomažin on 24/02/2020.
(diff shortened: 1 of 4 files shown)


Result+Povio.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Alamofire
+//
+//  Created by Borut Tomažin on 24/02/2020.
+//  Copyright © 2020 Povio Labs. All rights reserved.
+//
+
+import Foundation
+
+extension Result where Success == Void {
+  static func success() -> Result {
```
tonikocjan: Maybe computed prop instead?

borut-t: I would stick with the function instead. Since we have a function for non-void cases and also for failures, it would only bring confusion.

PR #65, opened 2020-05-12
Fix/SPM

borut-t:
Updated Package to work better with Swift 5.2. Resources and readme files are currently kept in a separate folder `Resources` because SPM doesn't support it yet. It will be available as part of [Swift 5.3](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md).
Another thing is that SPM currently doesn't handle well targets without dependencies. Therefore `Networking` is not part of SPM. Better support is also coming with Swift 5.3. If you need it, install it via cocoapods.


Package.swift
```diff
@@ -1,16 +1,23 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.1
 import PackageDescription
 
 let package = Package(
   name: "PovioKit",
   platforms: [
-    .macOS(.v10_14), .iOS(.v11), .tvOS(.v10), .watchOS(.v3)
+    .iOS(.v11),
+    .macOS(.v10_14),
```
tonikocjan, line 8: Should we disable `macOS` since we are using UIKit in several places?

borut-t: Good point. It was added when we were also using Alamofire dependency with SPM and it required macos platform. I'll remove it.

PR #66, opened 2020-05-12
Chore/Code updates

borut-t:
Updated code style and deprecated a few methods.


Sources/Extensions/Foundation/Collection+Povio.swift
```diff
@@ -11,14 +11,14 @@ import Foundation
 public extension Collection where Indices.Iterator.Element == Index {
   /// Returns the element at the specified `index` if it is within bounds, otherwise `nil`.
   subscript (safe index: Index) -> Iterator.Element? {
-    return (startIndex <= index && index < endIndex) ? self[index] : nil
+    (startIndex <= index && index < endIndex) ? self[index] : nil
```
tonikocjan, line 14: What about `indices.contains(index)` instead?

borut-t: `contains` is slow operation since it needs to loop over all elements in collection. This is especially evident in collections with lots of items.

PR #76, opened 2020-07-26
Bar button item

tonikocjan:
Simplify dealing with `UIBarButtonItem`s.

### DIFF
--- a/Sources/Extensions/UIKit/UIViewController+Povio.swift
+++ b/Sources/Extensions/UIKit/UIViewController+Povio.swift
@@ -8,44 +8,67 @@
 import UIKit
 
 public extension UIViewController {
-  struct BarButton {
-    let image: UIImage?
-    let title: Title?
+  class BarButton {
+    let content: Content
     let action: Selector
+    
+    init(content: Content, action: Selector) {
+      self.content = content
+      self.action = action
+    }
   }
   
-  func setLeftBarButton(_ barButton: BarButton) {
-    navigationItem.leftBarButtonItem = createButton(using: barButton)
+  enum Content {
+    case icon(UIImage)
+    case title(Title)
+    
+    static func icon(_ image: UIImage?) -> Content {
+      .icon(image ?? UIImage())
+    }
   }
   
-  func setRightBarButton(_ barButton: BarButton) {
-    navigationItem.rightBarButtonItem = createButton(using: barButton)
+  enum Title {
+    case `default`(String)
+    case attributed(normal: NSAttributedString, disabled: NSAttributedString?)
+    
+    static func attributed(normal: NSAttributedString) -> Title {
+      .attributed(normal: normal, disabled: nil)
+    }
   }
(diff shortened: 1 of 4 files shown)


Sources/Extensions/UIKit/UIViewController+Povio.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+//  Created by Toni Kocjan on 26/07/2020.
+//
+
+import UIKit
+
+public extension UIViewController {
+  class BarButton {
+    let content: Content
+    let action: Selector
```
nejcvivod, line 13: Should we add an optional `let target: Any?` here and then do `addTarget(barButton.target ?? self, ...`?
That way the caller can decide which object will receive the messages directly - edge case, so fine if we leave as-is 🤷

Sources/Extensions/UIKit/UIViewController+Povio.swift
```diff
@@ -0,0 +1,79 @@
...
+    let target: Any?
+    
+    required init(content: Content, action: Selector, target: Any? = nil) {
+      self.content = content
+      self.action = action
+      self.target = target
+    }
+  }
+  
+  enum Content {
```
borut-t, line 23: Let's wrap `Content` inside extension of `BarButton` class.

Sources/Extensions/UIKit/UIViewController+Povio.swift
```diff
@@ -0,0 +1,79 @@
...
+    case icon(UIImage)
+    case title(Title)
+    
+    static func icon(_ image: UIImage?) -> Content {
+      .icon(image ?? UIImage())
+    }
+  }
+  
+  enum Title {
```
borut-t, line 32: Let's wrap `Title` inside extension of `Content` enum.

Sources/Extensions/UIKit/UIViewController+Povio.swift
```diff
@@ -0,0 +1,79 @@
...
+      .attributed(normal: normal, disabled: nil)
+    }
+  }
+}
+
+extension UIViewController {
+  @discardableResult
+  func setLeftBarButton(_ barButton: BarButton) -> UIBarButtonItem {
+    let button = createButton(using: barButton)
+    navigationItem.leftBarButtonItem = button
```
borut-t, line 46: This will only leverage on bar button per left/right item. Should we pre-pend items to `leftBarButtonItems` array instead? Or maybe even accept `barButtons: [BarButton]` as param 🤷‍♂️

borut-t (commented):
Some thoughts and comments.

borut-t (approved):

borut-t:
@tonikocjan can you check the comments, make changes accordingly and merge this ticket?

tonikocjan:
I'm unable to successfully rebase this branch, so I'll close this ticket and open a new one.

PR #77, opened 2020-08-21
Feature/Hexadecimal data representation

borut-t:
Created a Data extension method which converts data object into a hexadecimal string.
One great example of its usage is for converting push notifications "deviceToken" data into a string.


nejcvivod (approved):
Lookin' good 👍

borut-t:
> @borut-t are any tests needed?

Good question. I think they are. I'll add them.

PR #81, opened 2020-10-27
Feature/Sign In with Apple

borut-t:


Sources/Utilities/SignInWithApple/SignInWithApple.swift
```diff
@@ -0,0 +1,200 @@
...
+      }
+      
+      randoms.forEach {
+        guard length != 0 else { return }
+        if $0 < charset.count {
+          result.append(charset[Int($0)])
+          remainingLength -= 1
+        }
+      }
+    }
```
nejcvivod, line 170: Hm, do we need to do this, or can we get away by using randomElement()? So something like:
```swift
  let result = (0..<length).compactMap { _ in charset.randomElement() }
  guard result.count == length else {
    fatalError(...)
  }
  return String(result)
}
```
Not sure what the distributions are like in both cases, but the default (SystemRandomNumberGenerator) generator should be fine (according to docs):

> SystemRandomNumberGenerator is automatically seeded, is safe to use in multiple threads, and uses a cryptographically secure algorithm whenever possible.

nejcvivod (approved):
👍 looks good, have one question regarding that random generator 🤔

Sources/Utilities/SignInWithApple/SignInWithApple.swift
borut-t: Never thought about this since the code was copy-pasted from Firebase [document](https://firebase.google.com/docs/auth/ios/apple). I'll take a look and see if we can make a replacement.

borut-t: Tested and it works. Thanks @nejcvivod 🙌

PR #83, opened 2020-11-09
Feature/ANC additions

borut-t:
Added a few things to ANC:
- AlamofireConsoleLogger - a separate event logger for the console; moving logging logic out from the client
- Request interceptor per request
- Generic retry interceptor


Sources/Networking/AlamofireNetworkClient/AlamofireConsoleLogger.swift
```diff
@@ -0,0 +1,46 @@
...
+  var method: String {
+    request?.method?.rawValue ?? "<no method>"
+  }
+  
+  var statusCode: Int {
+    response?.statusCode ?? 0
+  }
+  
+  var endpoint: String {
+    request?.url?.lastPathComponent ?? "<no endpoint>"
```
nejcvivod, line 44: Hm, might be handy to return everything after host - just in case we're in a situation where we have v1/user, v2/user endpoints (*sigh*)

That way we can skim the logs and see they're 2 different endpoints

nejcvivod (approved):
👍 looks great

Sources/Networking/AlamofireNetworkClient/AlamofireConsoleLogger.swift
borut-t: That's a great suggestion, smth we should definitely do.

borut-t:
@tonikocjan can I ask you for a review? This is smth that's based on your work.

tonikocjan (approved):
👌

Sources/Networking/AlamofireNetworkClient/AlamofireConsoleLogger.swift
borut-t: @nejcvivod https://github.com/poviolabs/PovioKit/pull/83/commits/c5dc4a25c6751a51cbc0e783a7b48bd5905ec9e1

PR #86, opened 2020-11-11
Feature/MapKit extensions

borut-t:
Added some useful MapKit extensions.

### DIFF
--- a/Sources/Extensions/MapKit/MKMapView+PovioKit.swift
+++ b/Sources/Extensions/MapKit/MKMapView+PovioKit.swift
@@ -9,15 +9,10 @@
 import MapKit.MKMapView
 
 public extension MKMapView {
-  /// Current zoom level
-  var zoom: Double {
-    log2(360 * ((Double(frame.size.width) / 128) / region.span.longitudeDelta))
-  }
-  
   /// Returns radius distance in meters from map center to the top of the visible screen
   var visibleRadius: Double {
     let centerLocation = CLLocation(latitude: centerCoordinate.latitude, longitude: centerCoordinate.longitude)
-    let topCenterCoordinate = convert(.init(x: bounds.size.width / 2.0, y: 0), toCoordinateFrom: self)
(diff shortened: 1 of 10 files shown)


Sources/Extensions/MapKit/MKMapView+PovioKit.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Borut Tomažin on 11/11/2020.
+//  Copyright © 2020 Povio Labs. All rights reserved.
+//
+
+import MapKit.MKMapView
+
+public extension MKMapView {
+  /// Current zoom level
+  var zoom: Double {
```
nejcvivod, line 13: What does this denote, percent?

Sources/Extensions/MapKit/MKMapView+PovioKit.swift
```diff
@@ -0,0 +1,42 @@
...
+    log2(360 * ((Double(frame.size.width) / 128) / region.span.longitudeDelta))
+  }
+  
+  /// Returns radius distance in meters from map center to the top of the visible screen
+  var visibleRadius: Double {
+    let centerLocation = CLLocation(latitude: centerCoordinate.latitude, longitude: centerCoordinate.longitude)
+    let topCenterCoordinate = convert(.init(x: bounds.size.width / 2.0, y: 0), toCoordinateFrom: self)
```
nejcvivod, line 20: We could use bounds.midX instead of bounds.size.width / 2.0 🤔

nejcvivod (approved):
👍 looks great

Sources/Extensions/MapKit/MKMapView+PovioKit.swift
borut-t: Good question. This was taken out of context, therefore not good enough to be put here. I'll remove it.

Sources/Extensions/MapKit/MKMapView+PovioKit.swift
borut-t: Good point 👌

PR #87, opened 2020-11-11
Feature/Foundation extensions

borut-t:
Some more Foundation extensions.


Sources/Extensions/Foundation/Double+Povio.swift
```diff
@@ -0,0 +1,24 @@
...
+  /// Returns miles in meters
+  var meters: Meters {
+    self * 1609.344
+  }
+  
+  /// Return meters in miles
+  var miles: Miles {
+    self / 1609.344
+  }
+}
```
nejcvivod, line 24: Not sure I like this, we can easily get into a situation where we do the same conversion twice :\ (meters to meters for example)

Should we have proper types wrapping this and then make conversions between those?

Sources/Extensions/Foundation/Dictionary+Povio.swift
```diff
@@ -0,0 +1,26 @@
...
+
+public extension Dictionary {
+  /// Make a union on two dictionaries.
+  /// ```
+  /// let dictionaryA = [1: "One", 2: "Two"]
+  /// let dictionaryB = [3: "Three"]
+  /// let unionDictionary = dictionaryA.union(dictionaryB)
+  /// print(unionDictionary) // [1: "One", 2: "Two", 3: "Three"]
+  /// ```
+  func union(_ dictionary: Dictionary) -> Dictionary {
```
tonikocjan, line 19: What happens if dictionaries contain the same key. Which value remains?

Sources/Extensions/Foundation/String+Povio.swift
```diff
@@ -37,4 +37,14 @@ public extension String {
...
+  
+  /// Returns first n characters from string
+  func first(n characters: Int) -> String {
+    .init(prefix(min(count, characters)))
+  }
+  
+  /// Returns last n characters from string
+  func last(n characters: Int) -> String {
+    .init(suffix(min(count, characters)))
+  }
```
tonikocjan, line 49: These seem redundant, `prefix` and `suffix` already do their job, do they not?

Sources/Extensions/Foundation/Dictionary+Povio.swift
borut-t: Original is preserved. Which is not good. I'll think about it.

Sources/Extensions/Foundation/Double+Povio.swift
borut-t: Thinking further about this if this makes sense at all. I'll think about it.

Sources/Extensions/Foundation/String+Povio.swift
borut-t: Yes, this is from pre-swift4 times.

Sources/Extensions/Foundation/Dictionary+Povio.swift
borut-t: Actually, this is redundant. [Dictionary](https://developer.apple.com/documentation/swift/dictionary/3127169-merge) already has `merge` and `merging` methods.

Sources/Extensions/Foundation/Double+Povio.swift
borut-t: @nejcvivod I've decided to remove it since it's not clear what it does with the current state.

PR #92, opened 2020-11-18
Fixes

tonikocjan:
1. File containing `BarButton` was missing.
2. Initializer for `OAuthContainer` must be public.


borut-t (approved):
Nice findings. Thanks for testing it out 🙌

PR #98, opened 2020-12-04
Feature/Network client modifications

borut-t:


Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -91,6 +92,33 @@ public extension AlamofireNetworkClient {
...
   }
   
+  func request<E: Encodable>(
+    method: HTTPMethod,
+    endpoint: URLConvertible,
+    headers: HTTPHeaders? = nil,
+    encode: E,
+    encoder: JSONEncoder? = nil,
+    interceptor: RequestInterceptor? = nil) -> Request
+  {
```
borut-t, line 102: @tonikocjan I've decided to go with `JSONEncoder` vs configuration here. We are doing the same on decoding end.

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -62,6 +62,7 @@ public extension AlamofireNetworkClient {
     return .init(with: request)
   }
   
+  @available(*, deprecated, message: "Use request method with `encoder` instead.")
   func request<E: Encodable>(
     method: HTTPMethod,
     endpoint: URLConvertible,
```
borut-t, line 68: Any objections to deprecating this method?

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -174,12 +202,9 @@ public extension AlamofireNetworkClient.Request {
     }
   }
   
-  func decode<D: Decodable>(
-    _ decodable: D.Type,
-    decoder: JSONDecoder) -> Promise<D>
-  {
+  func decode<D: Decodable>(_ decodable: D.Type, decoder: JSONDecoder? = nil) -> Promise<D> {
```
tonikocjan, line 205: Should we follow the same style throughout the file?

tonikocjan, line 205: Why optional decoder?

borut-t: We are not currently. There are some other methods that don't follow the same indentation rule. But I can make it if you prefer that way 🤷‍♂️

borut-t: Good question.

I think it makes sense to make it optional which can become handy if you just want to decode the response with the default rules.
On the other hand, making it as required, it's a bit wacky to create a new instance every time when you only need default rules.

What are your concerns?

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
borut-t: Actually, I need to delete this method since it has the same signature as the new method. Therefore compiler cannot tell which one to pick.

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
tonikocjan: You are creating a new instance inside anyway, right? So why not just
```
func decode<D: Decodable>(
    _ decodable: D.Type,
    decoder: JSONDecoder = .init()) -> Promise<D>
  {
```
?

borut-t: That's another way yes.

tonikocjan: I prefer this way because the dependency is not _hidden_ inside the method.

borut-t: Right. I've also changed the behavior for the encoder as well.

borut-t:
@tonikocjan please check https://github.com/poviolabs/PovioKit/pull/98/commits/fd7291efe85b5fa5069c55948cc4123560758dc6 also.

tonikocjan (approved):

PR #100, opened 2021-01-08
Fix/Padding label

borut-t:
Fixed issue where "text" size was not properly calculated.

### DIFF
--- a/Sources/Views/PaddingLabel/PaddingLabel.swift
+++ b/Sources/Views/PaddingLabel/PaddingLabel.swift
@@ -18,6 +18,10 @@ open class PaddingLabel: UILabel {
     super.drawText(in: rect.inset(by: contentInset))
   }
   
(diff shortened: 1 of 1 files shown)


Sources/Views/PaddingLabel/PaddingLabel.swift
```diff
@@ -18,6 +18,10 @@ open class PaddingLabel: UILabel {
     super.drawText(in: rect.inset(by: contentInset))
   }
   
+  override open func textRect(forBounds bounds: CGRect, limitedToNumberOfLines numberOfLines: Int) -> CGRect {
+    super.textRect(forBounds: bounds.inset(by: contentInset), limitedToNumberOfLines: 0)
```
EgzonArifi, line 22: can we pass .zero as a default value for `numberOfLines` instead?

borut-t: Tried. I doesn't work with that. Not sure why.

tonikocjan (approved):

PR #102, opened 2021-01-28
OAuthInterceptor - improvements

tonikocjan:
Added logic to check if `access_token` is valid at the point of refreshing which indicates that it was just refreshed so it's not necessary to refresh it again. Also, make sure that requests are not retried more than necessary.

### DIFF
--- a/Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
+++ b/Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
@@ -47,11 +47,6 @@ open class OAuthRequestInterceptor {
   private let lock: NSLock = .init()
   private var activeRequests: [Request: RequestState] = .init()
   
-  private enum RequestState {
-    case retry
-    case reject
-  }
-  
   public init(
     provider: OAuthProvider,
     storage: OAuthStorage,
@@ -60,6 +55,13 @@ open class OAuthRequestInterceptor {
     self.provider = provider
     self.storage = storage
     self.adapter = adapter
+  }
+}
+
+extension OAuthRequestInterceptor {
+  enum RequestState {
+    case retry
+    case reject
   }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
```diff
@@ -13,19 +13,23 @@ import PovioKit
 public struct OAuthContainer {
   public let accessToken: String
   public let refreshToken: String
+  public let expirationDate: Date
```
borut-t, line 16: Should this be optional? What if we don't have that mechanism defined?

borut-t (approved):
Great. One question though.

Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
tonikocjan: Applied some changes, please take a look again.

Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
```diff
@@ -13,27 +13,28 @@ import PovioKit
...
-  var accessTokenExpirationDate: Date? { get set }
 }
 
 public protocol OAuthProvider {
   func refresh(with refreshToken: String) -> Promise<OAuthContainer>
+  func isAccessTokenValid() -> Bool
+}
+
+public extension OAuthProvider {
+  func isAccessTokenValid() -> Bool { true }
```
borut-t, line 37: So if we don't override it, token is always "valid"? 🤔

tonikocjan: If we provide a default implementation then `false` doesn't make any sense. So we either remove the default implementation or keep it as is.
The reason for default implementation is that most of the time the app is not responsible to check whether the token is valid or not. That is the responsibility of the server.
If you actually have to perform some logic then it is up to you to implement it.

MrNiceMrGood (approved):

Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
tonikocjan: Actually, let me think a bit more about this. It might be best to just remove the default implementation and not rely on any assumptions.

borut-t (approved):
Like it. Ship it.

PR #113, opened 2021-07-29
Feature/PromiseKit - Package restructure + `chainIf` and `mapIf`

tonikocjan:
1. Implemented several additional functionalities to the PromiseKit:

- `chainIf` and `mapIf`, which enable the user to apply _non-determinism_ by splitting the chain into two new chains.
- Bunch of standard monad infix operators which are useful in some contexts (refer to given examples).

2. Re-structured the package so that `PromiseKit` is a separate sub-package.


Sources/PromiseKit/Promise.swift
```diff
@@ -766,3 +842,132 @@ extension AnyOptionalType where Self: OptionalType {
...
+///  .chain { f4($0.1.0) }
+///
+/// can be written as:
+///
+/// f1() >>- { v1 in
+///   f2() >>- { v2 in
+///     f3() >>- { v3 in
+///       f4(v1)
+///     }
+///   }
```
borut-t, line 912: I like the clarity of this approach but at the same time, I wonder if we want to do that. Isn't one of the selling points of promises to avoid nesting closures? 🤔

borut-t (approved):
I like improvements. Conditionals are rare but can happen and can be very helpful.

Sources/PromiseKit/Promise.swift
tonikocjan: I see your point and I agree, in most cases, one should avoid infix notation and use standard chaining. But in those other few cases (when you need a value located far up the chain) I see the value of the proposed approach.
On the other hand, I understand that for someone not familiar with such syntax it may look really strange at the first sight.

PR #115, opened 2021-07-29
Chore/ASB code update

tonikocjan:
Modernised and updated ASB code (note that this is a code-breaking change and will require, though minor, migration).


borut-t (commented):
Looks nice. Do we need to update the readme file as well?

PR #116, opened 2021-07-29
Chore/Demo App w/ ValidationForm examples

tonikocjan:
Prepared demo app with two examples of `ValidationForm`s.

all


Demo/Demo/Forms/FormsListViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+    let cell = UITableViewCell(style: .value1, reuseIdentifier: "cell")
+    cell.textLabel?.text = form.name
+    return cell
+  }
+}
+
+extension FormsListViewController: UITableViewDelegate {
+  func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+    let form = validationForms[indexPath.row].form
+    let vc = FormViewController(form: form)
```
ArlindDushi, line 50: As mentioned above, not sure about this, would prefer to  keep the current structure we use and have the form "embeded" into ViewController's `contentView` and not need to inject it.

Demo/Demo/Forms/FormViewController.swift
```diff
@@ -0,0 +1,44 @@
...
+import PovioKit
+
+class FormViewController<F: ValidationForm>: UIViewController {
+  let formView: ValidationFormView<F>
+  
+  init(form: F) {
+    formView = .init(validationForm: form)
+    super.init(nibName: nil, bundle: nil)
+    setupViews()
+  }
```
ArlindDushi, line 18: Maybe better to not inject the form directly, but have it create inside content view? Thus we keep the current structure we use in projects, what do you think 🤔 ?

ArlindDushi (commented):
Looks promising 👌 , please check my comments below.

Demo/Demo/Forms/FormsListViewController.swift
tonikocjan: This is just a demo showcasing how the validation from builder works. The way it is used in the demo is not meant to be translated into a real world project.

Demo/Demo/Forms/FormViewController.swift
tonikocjan: I agree that in the actual project you would not do it like this. But as this is just a demo, I don't think it matters much to be honest?

PR #119, opened 2021-08-26
Feature/PromiseKit - Renaming proposal + `Any` and `Race`

tonikocjan:
This ticket proposes the following naming changes:
  1. `combine` => `all`
  2. `onSuccess` => `then`
  3. `onFailure` => `catch`
  4. `observe` => `finally`

I think the proposed names better convey the underlying meaning and they more closely follow the _Promise standard_. The old methods were deprecated only, which means that existing code using PromiseKit should still compile.

Additionally, 
  - `any` (a dual of `all`), which allows the user to combine several promises allowing some to fail, and 
  - `race`, which returns a promise that fulfills or rejects as soon as one of the promises in the collection fulfills or rejects, with the result from that promise

are implemented (together with [tests](https://github.com/poviolabs/PovioKit/pull/119/files#diff-45eb96e7c84e64adf2ac03ce2fec2a6ad728250229a154f617b0c5358d7e3c57)).

### DIFF
--- a/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
+++ b/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
@@ -23,9 +23,14 @@ public typealias Writer = (String) -> Void
 
 open class AlamofireNetworkClient {
   private let session: Alamofire.Session
+  private let eventMonitors: [RequestMonitor]
   
-  public init(session: Alamofire.Session = .default) {
+  public init(
+    session: Alamofire.Session = .default,
+    eventMonitors: [RequestMonitor]
+  ) {
     self.session = session
+    self.eventMonitors = eventMonitors
   }
 }
 
@@ -42,7 +47,7 @@ public extension AlamofireNetworkClient {
                method: method,
                headers: headers,
                interceptor: interceptor)
-    return .init(with: request)
+    return .init(with: request, eventMonitors: eventMonitors)
   }
   
   func request(
@@ -60,7 +65,7 @@ public extension AlamofireNetworkClient {
                encoding: parameterEncoding,
                headers: headers,
                interceptor: interceptor)
-    return .init(with: request)
+    return .init(with: request, eventMonitors: eventMonitors)
   }
   
   func request<E: Encodable>(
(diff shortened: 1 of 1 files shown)


ArminSpahic (approved):

Sources/PromiseKit/Any.swift
```diff
@@ -0,0 +1,124 @@
...
+  guard !promises.isEmpty else {
+    return .error(NSError(domain: "com.poviokit.promisekit", code: 101, userInfo: nil))
+  }
+  
+  return .init { seal in
+    let barrier = DispatchQueue(label: "combineQueue", attributes: .concurrent)
+    for promise in promises {
+      promise.finally { result in
+        barrier.async(flags: .barrier) {
+          guard promises.allSatisfy({ $0.isResolved }) else { return }
```
borut-t, line 31: `isResolved` flags what?

Sources/PromiseKit/Race.swift
```diff
@@ -0,0 +1,41 @@
...
+public func race<T, C: Collection>(
+  on dispatchQueue: DispatchQueue = .main,
+  promises: C
+) -> Promise<T> where C.Element == Promise<T> {
+  guard !promises.isEmpty else {
+    return .error(NSError(domain: "com.poviokit.promisekit", code: 101, userInfo: nil))
+  }
+  
+  return .init { seal in
+    let barrier = DispatchQueue(label: "combineQueue", attributes: .concurrent)
```
borut-t, line 25: Why do you name queue "barrier" 🤔

borut-t (approved):
Great work and I think renaming following the Promise naming convention makes sense. Can you please update documentation and add examples for added options?

Sources/PromiseKit/Any.swift
tonikocjan: `isResolved` means is that the promise was fulfilled with a value.

Sources/PromiseKit/Race.swift
tonikocjan: https://developer.apple.com/documentation/dispatch/dispatch_barrier

tonikocjan:
>  Can you please update documentation and add examples for added options?

Absolutely.

PR #120, opened 2021-09-20
Feature/NotificationCenterDispatcher

borut-t:
`NotificationCenterDispatcher` helper for sending/receiving NotificationCenter broadcast messages.


Sources/Utilities/NotificationCenterDispatcher/KeyboardNotification.swift
```diff
@@ -0,0 +1,44 @@
...
+//  KeyboardNotification.swift
+//  PovioKit
+//
+//  Created by Borut Tomazin on 20/09/2021.
+//  Copyright © 2021 Povio Labs. All rights reserved.
+//
+
+import UIKit
+
+struct KeyboardNotification {
```
tonikocjan, line 11: If you want to expose this struct to the outside world it needs to be marked `public`; the same for all properties.

Sources/Utilities/NotificationCenterDispatcher/KeyboardNotification.swift
```diff
@@ -0,0 +1,44 @@
...
+    (userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue ?? .zero
+  }
+  
+  init(notification: Notification) {
+    self.notification = notification
+    if let userInfo = (notification as NSNotification).userInfo {
+      self.userInfo = userInfo as NSDictionary
+    } else {
+      self.userInfo = NSDictionary()
+    }
```
tonikocjan, line 42: Why not just parse everything here instead of doing it each time within computed properties?

Sources/Utilities/NotificationCenterDispatcher/KeyboardNotification.swift
borut-t: Totally forget about this. Thank 👌

Sources/Utilities/NotificationCenterDispatcher/NotificationCenterDispatcher.swift
```diff
@@ -0,0 +1,75 @@
...
+//  NotificationCenterDispatcher.swift
+//  PovioKit
+//
+//  Created by Borut Tomazin on 20/09/2021.
+//  Copyright © 2021 Povio Labs. All rights reserved.
+//
+
+import UIKit
+
+enum NotificationCenterDispatcher {
```
tonikocjan, line 11: Needs to be marked as `public`.

Sources/Utilities/NotificationCenterDispatcher/KeyboardNotification.swift
borut-t: Good point. Idk why I did it this way. I'll think about if there are any implications on that.

Sources/Utilities/NotificationCenterDispatcher/KeyboardNotification.swift
```diff
@@ -8,37 +8,26 @@
...
-    (userInfo[UIResponder.keyboardFrameBeginUserInfoKey] as? NSValue)?.cgRectValue ?? .zero
-  }
-  var keyboardFrameEnd: CGRect {
-    (userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue ?? .zero
-  }
+  let keyboardSize: CGSize
+  let keyboardFrameBegin: CGRect
+  let keyboardFrameEnd: CGRect
+  let animationDuration: TimeInterval
+  let animationCurve: UIView.AnimationOptions
```
tonikocjan, line 17: These should probably also be public.

PR #122, opened 2021-09-23
Feature/PromiseKit - Concurrent dispatch

tonikocjan:
Implemented an abstraction that provides a simple way of executing tasks concurrently. An example use case is to upload a file to a server by chunking it and uploading several chunks at the same time.


borut-t (approved):
Nice. Can we add an example to the README file as well?

PR #123, opened 2021-09-23
Feature/PromiseKit - Polling

tonikocjan:
Implemented `polling` abstraction which is useful in scenarios where you have to repeatedly (with some delay) do some work (query the server whether the status is ok for example).


borut-t (approved):
Great work 🙌 
Can we also update the README with an example?

PR #127, opened 2021-10-28
ANC: Monitor progress

tonikocjan:
Added capability to observe the upload and/or download progress of a request.

The addition should not break the existing code.


borut-t (approved):
👌

PR #128, opened 2021-10-29
AudioVideoPlayer

tonikocjan:
An `AVPlayer` wrapper with additional functionality like:
  1. looping
  2. state (`isPlaying`)
  2. observing different events
  3. easier playback intervals manipulation,
  4. ...

### DIFF
--- a/Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
+++ b/Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
@@ -1,0 +1,157 @@
+//
+//  AudioVideoPlayer.swift
+//  PovioKit
+//
+//  Created by Toni Kocjan on 29/10/2021.
+//  Copyright © 2021 Povio Labs. All rights reserved.
+//
+
+import AVKit
+
+public protocol AudioVideoPlayerDelegate: AnyObject {
+  func audioVideoPlayer(didBeginPlayback player: AudioVideoPlayer)
+  func audioVideoPlayer(didEndPlayback player: AudioVideoPlayer)
+  func audioVideoPlayer(didPausePlayback player: AudioVideoPlayer)
+  func audioVideoPlayer(didBeginReplay player: AudioVideoPlayer)
+  func audioVideoPlayer(didBeginBuffering player: AudioVideoPlayer)
+  func audioVideoPlayer(didEndBuffering player: AudioVideoPlayer)
+  func audioVideoPlayer(_ player: AudioVideoPlayer, didProgressToTime seconds: Double)
+}
+
+public extension AudioVideoPlayerDelegate {
+  func audioVideoPlayer(didBeginPlayback player: AudioVideoPlayer) {}
+  func audioVideoPlayer(didEndPlayback player: AudioVideoPlayer) {}
+  func audioVideoPlayer(didPausePlayback player: AudioVideoPlayer) {}
+  func audioVideoPlayer(didBeginReplay player: AudioVideoPlayer) {}
+  func audioVideoPlayer(didBeginBuffering player: AudioVideoPlayer) {}
+  func audioVideoPlayer(didEndBuffering player: AudioVideoPlayer) {}
+  func audioVideoPlayer(_ player: AudioVideoPlayer, didProgressToTime seconds: Double) {}
+}
+
+public class AudioVideoPlayer: AVPlayer {
+  public static let defaultTimescale = CMTimeScale(kCMTimeMaxTimescale)
+
+  public private(set) lazy var playbackInterval: (startAt: Double, endAt: Double) = (0, duration)
+  public var allowLooping = true
+  public var periodicTimeObserverTimeInterval: CMTime = CMTimeMake(value: 1, timescale: AudioVideoPlayer.defaultTimescale)
+  public weak var delegate: AudioVideoPlayerDelegate?
+  
+  private var timeObserver: Any?
+
+  public override init() {
+    super.init()
+    setupTimeObserver()
+  }
+
+  public override init(url: URL) {
+    super.init(url: url)
+    setupTimeObserver()
+  }
+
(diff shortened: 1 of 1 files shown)


MrNiceMrGood (approved):
Looks good to me

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
```diff
@@ -0,0 +1,154 @@
...
+    guard let url = URL(string: urlString) else { return nil }
+    self.init(url: url)
+  }
+}
+
+// MARK: - API
+public extension AudioVideoPlayer {
+  override func play() {
+    super.play()
+    isPlaying = true
```
borut-t, line 66: Do we really need this flag? I would rather go with [rate](https://developer.apple.com/documentation/avfoundation/avplayer/1388846-rate?preferredLanguage=occ) instead. What do you think?
```swift
var isPlaying: Bool {
  rate != 0
}
```

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
```diff
@@ -0,0 +1,154 @@
...
+    currentItem?.asset.duration.seconds ?? 0
+  }
+}
+
+// MARK: - Private Methods
+private extension AudioVideoPlayer {
+  func setupTimeObserver() {
+    guard timeObserver == nil else { return }
+
+    let timeInterval = CMTimeMake(value: 1, timescale: 10)
```
borut-t, line 120: So the observer will trigger every 1/10th of a second. What if we need it to be triggered more/less often? Should we make this configurable?

borut-t (commented):
I like it. Just added two comments to think about 🤔

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
tonikocjan: This looks very promising, yes. 👍  Will test it out if it works as expected.

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
tonikocjan: We should, of course.

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
```diff
@@ -0,0 +1,169 @@
...
+    setupTimeObserver()
+  }
+
+  var duration: Double {
+    currentItem?.asset.duration.seconds ?? 0
+  }
+  
+  var isPlaying: Bool {
+    rate != 0.0
+  }
```
gentique, line 127: It seems that it's possible for the player to stall due to buffer issues, etc. In that case the rate seems to not go to 0. We should also check if there's any error.

```suggestion
  var isPlaying: Bool {
    rate != 0.0 && error == nil
  }
```

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
```diff
@@ -0,0 +1,169 @@
...
+    playbackInterval = (startAt, endAt)
+  }
+
+  func setPlaybackPosition(
+    startAt: Double,
+    timescale: CMTimeScale = AudioVideoPlayer.defaultTimescale
+  ) {
+    removeTimeObserver()
+    seek(to: CMTime(seconds: startAt, preferredTimescale: timescale)) { _ in
+      self.setupTimeObserver()
```
gentique, line 109: Is it possible to have a retain cycle here?

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
tonikocjan: Used `status` instead, but very good point.

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
tonikocjan: The completion handler is called when `seek` either succeeds or is interrupted, so I think not.

gentique (approved):

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
```diff
@@ -0,0 +1,169 @@
...
+    removeTimeObserver()
+    setupTimeObserver()
+  }
+
+  /// Set `playbackInterval` and force player to start playing from the `startAt` timestamp.
+  func setPlaybackInterval(
+    startAt: Double,
+    endAt: Double,
+    timescale: CMTimeScale = AudioVideoPlayer.defaultTimescale
+  ) {
```
borut-t, line 97: Please rename to `play` and update the comment to
```swift
/// Starts and ends playing at `startAt` and `endAt` respectively.
```

Sources/Utilities/AudioVideoPlayer/AudioVideoPlayer.swift
```diff
@@ -0,0 +1,169 @@
...
+    guard startAt < endAt else { fatalError("`startAt` should be less than `endAt`") }
+    setPlaybackPosition(startAt: startAt)
+    playbackInterval = (startAt, endAt)
+  }
+
+  func setPlaybackPosition(
+    startAt: Double,
+    timescale: CMTimeScale = AudioVideoPlayer.defaultTimescale
+  ) {
```
borut-t, line 106: Please rename to `play` and add a comment
```swift
/// Starts playing at `startAt`.
```

borut-t (approved):
Two notes and we are ready.

tonikocjan:
I'm closing this one for the time being. Let's maybe consider it again in the future.

PR #130, opened 2021-11-04
Feature/LoadingView and LoadingBarButtons

tonikocjan:


Sources/Views/LoadingView.swift
```diff
@@ -0,0 +1,100 @@
...
+//  Copyright © 2021 Povio Labs. All rights reserved.
+//
+
+import UIKit
+
+/// Adapt any view into a loading view.
+open class LoadingView<V: UIView>: UIControl {
+  public let view: V
+  public let activityIndicator: UIActivityIndicatorView
+  let animations: Animations?
```
borut-t, line 15: This seems like overkill, nesting views. Could we just make a protocol and a method that will start/stop the animation? At start, we could insert the indicator, on the stop we should remove it 🤷‍♂️ 🤔

tonikocjan: I would argue that adding (together with layout-ing) and removing a subview is much more expensive than just hiding a view.
Not sure how you would solve this with a protocol. Can you give an example?

borut-t: What you are currently doing here is wrapping a "view" with another view, creating an `activityIndicator` inside and maybe all those things will not be needed at all in the lifecycle of the view.

What I would do is to just add `activityIndicator` to the view when needed, via extension method, smth like that:
```swift
extension UIView {
  func showLoader() {
    // add subview
    // set constraints
    // set tag
    // start loading
  }

  func hideLoader() {
    // find subview by tag
    // stop loading
    // remove
  }
}
```
Sure, this approach is a bit unreliable since we are relying on the `tag` which can be easily overridden. But I just wanted to present the idea, how this "loader" thing could be lightweight and only created when needed.
What do you think?

tonikocjan: Doing it in such a way, wouldn't the activity indicator cover the content?

image

borut-t: Isn't that how it suppose to be?

tonikocjan: At least the label should be hidden, I think. Otherwise, it looks kind of silly. 🤷 

https://user-images.githubusercontent.com/15092411/141788708-d6bd6b08-fe38-4978-a5ae-a532fcda1fe0.mov

With this implementation you can easily add an activity indicator to an image view with a nice fade animation:

https://user-images.githubusercontent.com/15092411/141788775-93fbe5ac-a19a-419e-b8be-c12cebb34fe6.mov

_p.s. I don't have a strong opinion on this ticket. We can go with the suggested lightweight implementation instead._

borut-t: Good point 👌 
Neither do I. Let's wait for other guys opinions.

Sources/Views/LoadingView.swift
```diff
@@ -0,0 +1,100 @@
...
+  
+  public init(
+    adapt view: V,
+    activityIndicatorStyle style: UIActivityIndicatorView.Style = .white,
```
gentique, line 19: We should use `.medium` here since .`white` is going to be deprecated.

Sources/Views/LoadingView.swift
gentique: I tried to use this in a simple case, where I wanted to show an activity indicator over the whole screen e.g. `self.view`, It did give me some unexpected results (screen went black), and also as @borut-t mentioned the nesting when used incorrectly (such as in my case with multiple ui elements) could produce a negative performance impact.

Sources/Views/LoadingView.swift
borut-t: Good point 👌

ArlindDushi (changes requested):
I tried adding a `leftBarButtonItem` but it's not appearing, not sure if I'm doing something wrong or maybe the implementation isn't working properly 🤷‍♂️  

Can we please have some simple examples on how to use these?

markom-povio:
I agree with previous comments, adding examples could be really helpful
I have tried to use it, but only managed to trigger start/stop animation closure (I needed to change view's background to track) but loader was not visible on the screen :/ 
It seems like a good idea to use in this special cases like in the PR code (adding loader to the BarButton) but I did not manage to make it work...

PR #131, opened 2021-11-19
Feature/Add speed and direction to UIView rotate func

gentique:
-Added default params for speed and direction.
-Reworked logic to support new params.

### DIFF
--- a/Sources/Extensions/UIKit/UIView+Povio.swift
+++ b/Sources/Extensions/UIKit/UIView+Povio.swift
@@ -59,7 +59,7 @@ public extension UIView {
     guard layer.animation(forKey: UIView.AnimationKey.rotation) == nil else { return }
     let rotationAnimation = CABasicAnimation(keyPath: "transform.rotation")
     rotationAnimation.fromValue = 0.0
-    rotationAnimation.toValue = (Float.pi * 2.0) * (clockWiseDirection ? 1 : -1)
+    rotationAnimation.toValue = (Float.pi * 2.0) * (clockwise ? 1 : -1)
     rotationAnimation.duration = speed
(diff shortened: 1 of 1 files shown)


Sources/Extensions/UIKit/UIView+Povio.swift
```diff
@@ -55,12 +55,12 @@ public extension UIView {
     static let shadowOpacity = "shadowOpacityKey"
   }
   
-  func rotate() {
+  func rotate(speed: CFTimeInterval = 1.25, clockWiseDirection: Bool = true) {
```
tonikocjan, line 58: I think `clockWiseDirection` can be shortened to just `clockwise'.

Sources/Extensions/UIKit/UIView+Povio.swift
```diff
@@ -55,12 +55,12 @@ public extension UIView {
...
     if layer.animation(forKey: UIView.AnimationKey.rotation) == nil {
```
tonikocjan, line 59: Not necessarily part of this ticket, but can we use guard instead?

tonikocjan (approved):

Sources/Extensions/UIKit/UIView+Povio.swift
borut-t: Good one 👍

borut-t (approved):

Sources/Extensions/UIKit/UIView+Povio.swift
gentique: @tonikocjan thanks for the suggestion, we can remove 'direction' since it can be implied from the name of the function.

PR #133, opened 2021-11-22
Feature/PromiseKit - API Update

tonikocjan:
I propose the following name change:
`chain` -> `flatMap`

The reasoning is that `flatMap` matches the naming of Swift types that belong to the same family of types as `Promise`. These, among others, are: `Array` and other Collections, `Optional`, ...

By making this change I want to make it more clear when `flatMap` (i.e. `chain`) should be used instead of `map`; that is, in the same situations you would use `flatMap` for lists and optionals.


borut-t (approved):
👌

Aragunz (approved):

PR #138, opened 2022-01-20
Chore/Remove unnecessary log check

gentique:
Log level was checked twice.


ArlindDushi (approved):
Looks good to me. We do this check in `flush` method anyway.

tonikocjan (approved):
👍

borut-t (approved):
Nice catch 👌

PR #139, opened 2022-01-24
Feature/Textfield

ArlindDushi:
A generic-customizable textfield with optional validation.

From left to right: 
Normal textfield, textfield with title, textfield if validation fails

screenshots

Included `RuleValidatable` protocol to validate input and show appropriate error msg.

Usage:  
```
struct RequiredRule: RuleValidatable {
  let error: String
  
  init(error: String = "Need input") {
    self.error = error
  }
  
  func validate(_ input: String?) -> Bool {
    input.noneIfEmpty != .none
  }
}
```
.
.
.

```
let requiredRule = RequiredRule()
let validatableTextField = TextField(with: requiredRule)
```

Validation will be triggered when `validatableTextField.isValid` is called and if validation doesn't pass then the error message will automatically be shown in the bottom of textfield.

Tried to include a lot of properties but if you notice something missing please let me know.


Sources/Core/Views/TextField/TextField.swift
```diff
@@ -0,0 +1,304 @@
...
+  public var inputFieldCornerRadius: CGFloat = 10 {
+    didSet { valueContainerView.layer.cornerRadius = inputFieldCornerRadius }
+  }
+  
+  private var error: String? {
+    didSet { errorStateChanged() }
+  }
+  
+  public var hideKeyboardOnReturn = false
+  public var shouldChangeCharacters: Bool = true
```
gentique, line 66: Could you add more context as to what this bool does, and also maybe provide a way to add text validation for `textField(_:shouldChangeCharactersIn:replacementString:)`

gentique (changes requested):
LGTM, just some minor suggestions. Thought if we want to include this, we should also add it in the main Readme file and add some documentation with it.

Sources/Core/Views/TextField/TextField.swift
```diff
@@ -0,0 +1,304 @@
...
+  }
+  
+  public var value: String? {
+    get {
+      return valueTextField.text
+    }
+    set {
+      valueTextField.text = newValue
+    }
+  }
```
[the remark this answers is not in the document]
ArlindDushi: Yes, I agree. Will change this.

Sources/Core/Views/TextField/TextField.swift
ArlindDushi: After rethinking this I think offering a protocol to the implementer to do its own implementation is a better way to go. Will do the necessary changes here.

Sources/Core/Views/TextField/TextField.swift
```diff
@@ -0,0 +1,272 @@
+//
+//  TextField.swift
+//  
+//
+//  Created by Arlind Dushi on 01/20/2022.
+//
```
gentique, line 6: Please add `Povio Inc` in the signature.

gentique (commented):
Except for the header signature, LGTM.

Though seeing lots of wrappers for UITextField's properties makes this class high maintenance in the sense that when in the future a new property is added or needed we need to also change the code here.

This got me thinking and I'm wondering if we could inherit from UITextField instead of UIView, that way we would have all of the properties with the benefit of added functionality.

ArlindDushi:
> Except for the header signature, LGTM.
> 
> Though seeing lots of wrappers for UITextField's properties makes this class high maintenance in the sense that when in the future a new property is added or needed (to UITextField) we need to also change the code here.
> 
> This got me thinking and I'm wondering if we could inherit from UITextField instead of UIView, that way we would have all of the properties with the benefit of added functionality.

I get your point but wouldn't you lose the ability to show title and error label? This is not simply a lone UITextField being shown but a structure of UI components bundled together that give some ready-made capabilities to the implementer.

ArlindDushi:
> Except for the header signature, LGTM.
> 
> Though seeing lots of wrappers for UITextField's properties makes this class high maintenance in the sense that when in the future a new property is added or needed (to UITextField) we need to also change the code here.
> 
> This got me thinking and I'm wondering if we could inherit from UITextField instead of UIView, that way we would have all of the properties with the benefit of added functionality.

I get your point, but wouldn't we lo

> Except for the header signature, LGTM.
> 
> Though seeing lots of wrappers for UITextField's properties makes this class high maintenance in the sense that when in the future a new property is added or needed (to UITextField) we need to also change the code here.
> 
> This got me thinking and I'm wondering if we could inherit from UITextField instead of UIView, that way we would have all of the properties with the benefit of added functionality.

Fair point but we have some additional UI components bundled together here and in that case wouldn't we lose the ability to show error & title label?

gentique (approved):

Sources/Core/Views/TextField/TextField.swift
```diff
@@ -0,0 +1,273 @@
...
+  
+  override public var isFirstResponder: Bool {
+    valueTextField.isFirstResponder
+  }
+  
+  public var inputFieldHeight: CGFloat = 44 {
+    didSet { textFieldHeightAnchor?.constant = inputFieldHeight }
+  }
+  
+  public var inputFieldCornerRadius: CGFloat = 10 {
```
markom-povio, line 57: This is duplicated with _textFieldCornerRadius_
I guess the name textFieldCornerRadius is more appropriate in this case to stay consistent with the rest of the properties...

Sources/Core/Views/TextField/TextField.swift
```diff
@@ -0,0 +1,273 @@
...
+  private let valueTextField = UITextField()
+  private let errorLabel = UILabel()
+  private var rule: RuleValidatable?
+  private var textFieldHeightAnchor: NSLayoutConstraint?
+
+  weak public var delegate: UITextFieldDelegate? {
+    didSet { valueTextField.delegate = delegate }
+  }
+  
+  public init(with rule: RuleValidatable? = .none) {
```
markom-povio, line 29: Maybe it would be also good to have option to set rule properly separate from init

`func setRule(_ rule: RuleValidatable) {`

markom-povio (approved):
Looking good 👍

Resources/Views/TextField/README.md
```diff
@@ -0,0 +1,37 @@
+#  TextField
```
borut-t, line 1: Let's move this file into `Resources/UI/TextField/` folder.

Sources/Core/Views/TextField/TextField.swift
```diff
@@ -0,0 +1,276 @@
+//
+//  TextField.swift
```
borut-t, line 2: Let's move this class into `Sources/UI/TextField/` folder instead.

borut-t (approved):
Great work. Please rebase and fix suggested changed.

ArlindDushi:
Hey @gentique & @tonikocjan 

As noted [here](https://github.com/poviolabs/PovioKit/pull/178#issuecomment-1205019874) by @borut-t the shorter delays introduced in that PR caused `PromiseTests.testAllResolvedOnMainThread()` to fail and as a consequence this PR  could not be merged b/c of that.

Since the tests are passing with the previous value I have reverted to that value for now as to not block the merging.

gentique:
I tried running the tests repeatedly and couldn't get them to fail, none the less, I suggest that we add the delay only for `testAllResolvedOnMainThread()` as an initial workaround until this is fully resolved. If the other tests perform as expected we shouldn't burden them with the additional delay.

``` swift
  func testAllResolvedOnMainThread() {
    let ex1 = expectation(description: "")
    let ex2 = expectation(description: "")
    all(on: .main, promises: [async(0, delay: 0.05 ,on: .main), async(1, delay: 0.05, on: .main)])
      .then { _ in
        XCTAssertTrue(Thread.isMainThread)
        ex1.fulfill()
      }
    all(on: .main, promises: [async(0, on: .main), async(NSError.err, Int.self, delay: 0.05, on: .main)])
      .catch { _ in
        XCTAssertTrue(Thread.isMainThread)
        ex2.fulfill()
      }
    waitForExpectations(timeout: 2)
  }
```

I did some further evaluation and for example `testAllListAsyncFailing` if we set the delay to 0. `func all<T, C>(on dispatchQueue: DispatchQueue? = .main, promises: C)` fails to stop on the first error promise. I don't want to hijack this PR with these discussions but I believe this needs evaluation since it's not performing as intended.

Package.swift
```diff
@@ -37,7 +37,9 @@ let package = Package(
     ),
     .target(
       name: "PovioKitUI",
-      dependencies: [],
+      dependencies: [
+        "PovioKit"
+      ],
```
borut-t, line 42: Why do we need this dependency?

ArlindDushi: I'm using `UIColor` extension and `Builder` class of `PovioKit`. In fact I started using them before we split the UI component on its own. 

We can replace these auxiliary with "system" methods for both, to get rid of the dependency, what do you think?

ArlindDushi:
> I tried running the tests repeatedly and couldn't get them to fail, none the less, I suggest that we add the delay only for `testAllResolvedOnMainThread()` as an initial workaround until this is fully resolved. If the other tests perform as expected we shouldn't burden them with the additional delay.
> 
> ```swift
>   func testAllResolvedOnMainThread() {
>     let ex1 = expectation(description: "")
>     let ex2 = expectation(description: "")
>     all(on: .main, promises: [async(0, delay: 0.05 ,on: .main), async(1, delay: 0.05, on: .main)])
>       .then { _ in
>         XCTAssertTrue(Thread.isMainThread)
>         ex1.fulfill()
>       }
>     all(on: .main, promises: [async(0, on: .main), async(NSError.err, Int.self, delay: 0.05, on: .main)])
>       .catch { _ in
>         XCTAssertTrue(Thread.isMainThread)
>         ex2.fulfill()
>       }
>     waitForExpectations(timeout: 2)
>   }
> ```
> 
> I did some further evaluation and for example `testAllListAsyncFailing` if we set the delay to 0. `func all<T, C>(on dispatchQueue: DispatchQueue? = .main, promises: C)` fails to stop on the first error promise. I don't want to hijack this PR with these discussions but I believe this needs evaluation since it's not performing as intended.

Test didn't fail on my machine, it failed in the CI when I pushed a new commit. I had to revert the value for the test in CI to pass.

gentique:
> > I tried running the tests repeatedly and couldn't get them to fail, none the less, I suggest that we add the delay only for `testAllResolvedOnMainThread()` as an initial workaround until this is fully resolved. If the other tests perform as expected we shouldn't burden them with the additional delay.
> > ```swift
> >   func testAllResolvedOnMainThread() {
> >     let ex1 = expectation(description: "")
> >     let ex2 = expectation(description: "")
> >     all(on: .main, promises: [async(0, delay: 0.05 ,on: .main), async(1, delay: 0.05, on: .main)])
> >       .then { _ in
> >         XCTAssertTrue(Thread.isMainThread)
> >         ex1.fulfill()
> >       }
> >     all(on: .main, promises: [async(0, on: .main), async(NSError.err, Int.self, delay: 0.05, on: .main)])
> >       .catch { _ in
> >         XCTAssertTrue(Thread.isMainThread)
> >         ex2.fulfill()
> >       }
> >     waitForExpectations(timeout: 2)
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I did some further evaluation and for example `testAllListAsyncFailing` if we set the delay to 0. `func all<T, C>(on dispatchQueue: DispatchQueue? = .main, promises: C)` fails to stop on the first error promise. I don't want to hijack this PR with these discussions but I believe this needs evaluation since it's not performing as intended.
> 
> Test didn't fail on my machine, it failed in the CI when I pushed a new commit. I had to revert the value for the test in CI to pass.

Let's drop that commit and rebase to develop after #184 and #180 are merged. Thanks for the speedy fix @tonikocjan 🙌

Package.swift
borut-t: Ah ok. That's fine. It would be a strange case to only use `PovioKitUI` in the project.

PR #143, opened 2022-03-03
Register cells using variadic arguments

yllfejziu:
Current implementation allows you to register only one cell at a time.
This PR makes it possible to pass variadic arguments to the new added methods, registering more than one cell at a time.

Old implementation:

```
tableView.register(cell: ContactsSourceCell.self)
tableView.register(cell: ContactsSyncSourcesCell.self)
```

New implementation:

```
tableView.register(
    ContactsSourceCell.self,
    ContactsSyncSourcesCell.self
 )
```

### DIFF
--- a/Sources/Core/Extensions/UIKit/UICollectionView+Povio.swift
+++ b/Sources/Core/Extensions/UIKit/UICollectionView+Povio.swift
@@ -9,20 +9,20 @@
 import UIKit
 
 public extension UICollectionView {    
-  func register<T: UICollectionViewCell>(_: T.Type) {
-    register(T.self, forCellWithReuseIdentifier: T.identifier)
+  func registerCells<T: UICollectionViewCell>(_ cells: T.Type...) {
+    cells.forEach { register($0.self, forCellWithReuseIdentifier: $0.identifier) }
   }
   
   func registerSupplementaryView<T: UICollectionReusableView>(_: T.Type, kind: String) {
     register(T.self, forSupplementaryViewOfKind: kind, withReuseIdentifier: T.identifier)
   }
   
-  func register<T: UICollectionReusableView>(headerView: T.Type) {
-    registerSupplementaryView(headerView, kind: UICollectionView.elementKindSectionHeader)
+  func registerHeaderViews<T: UICollectionReusableView>(_ headerViews: T.Type...) {
+    headerViews.forEach { registerSupplementaryView($0, kind: UICollectionView.elementKindSectionHeader) }
   }
   
-  func register<T: UICollectionReusableView>(footerView: T.Type) {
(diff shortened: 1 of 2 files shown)


borut-t (approved):
I like the change. One comment though about the implementation: Put back deprecated method where they were initially, creating a separate code block doesn't make sense imo.

gentique (approved):

tonikocjan:
I like the general idea. I'm thinking about if it would be better to have all methods named simply `register`:

```swift
  func register<T: UICollectionViewCell>(_ cells: T.Type...)
  func register<T: UICollectionReusableView>(_: T.Type, kind: String) 
  func register<T: UICollectionReusableView>(_ headerViews: T.Type...)
  func register<T: UICollectionReusableView>(_ footerViews: T.Type...)
```

This makes it a little bit more concise IMO. What do you think?

Second, I'm not sure if the implementation is correct. By having a generic parameter you are basically saying that all cells have to be of the same type `T`. However, in general, you want to register cells of different types. Or am I missing something?

I think getting rid of the template parameter would solve the problem, i.e.:

```swift
extension UITableView {
  func register(_ cells: UITableViewCell.Type...) {
    cells.forEach { register($0.self, forCellReuseIdentifier: $0.identifier) }
  }
}
```

yllfejziu:
> I like the change. One comment though about the implementation: Put back deprecated method where they were initially, creating a separate code block doesn't make sense imo.

I also had a doubt when doing this, the main purpose of grouping deprecated methods was so we could identify them quicker and remove them all at once when needed. 
I put them back as they were for now.

yllfejziu:
> I like the general idea. I'm thinking about if it would be better to have all methods named simply `register`:
> 
> ```swift
>   func register<T: UICollectionViewCell>(_ cells: T.Type...)
>   func register<T: UICollectionReusableView>(_: T.Type, kind: String) 
>   func register<T: UICollectionReusableView>(_ headerViews: T.Type...)
>   func register<T: UICollectionReusableView>(_ footerViews: T.Type...)
> ```
> 
> This makes it a little bit more concise IMO. What do you think?
> 
> Second, I'm not sure if the implementation is correct. By having a generic parameter you are basically saying that all cells have to be of the same type `T`. However, in general, you want to register cells of different types. Or am I missing something?
> 
> I think getting rid of the template parameter would solve the problem, i.e.:
> 
> ```swift
> extension UITableView {
>   func register(_ cells: UITableViewCell.Type...) {
>     cells.forEach { register($0.self, forCellReuseIdentifier: $0.identifier) }
>   }
> }
> ```

Good catch on the generic parameter.
Also for the naming part, I've renamed functions that I could to 'register' only.

borut-t:
@yllfejziu looks like we can merge this right?

Issue #145, opened 2022-03-08
[UI] Profile ImageView


borut-t:
I would like to have a component for showing up user profile picture or initials that we often use on user profile pages or on user lists. It should be generic and customizable by (and not limited to) the following parameters:
- border
  - width
  - color
  - radius (an enum to define perfect circle or custom value)
- background color
  - plain color
  - gradient color
- placeholder image
  - when set, should be put on top of the background
- profile picture and initials label
  - a setter for both
  - if both are set, the last setter takes over the UI
  - contentMode should be set to `scaleAspectFill` and cannot be changed?
- badge view
  - can be added into any of the corner
  - circle shape
  - border color and width
  - filler picture
  - background color
- tap event handler using closure
- written in SwiftUI
- an example how to use it in UIKit and SwiftUI environment

PR #148, opened 2022-03-14
Feature/Profile image view

ArlindDushi:
https://github.com/poviolabs/PovioKit/issues/145

UI component written in SwiftUI. Can be imported into UIKit as well (iOS 13+) 
As this is the first time for me digging into SwiftUI please feel free to add any suggestions regarding the implementation or if you think there's a better way to do something.

Example on the left with placeholder and on the right when image is downloaded from url
merged

### Example: Implementation in SwiftUI
```
struct ContentView: View {
  var profileImageView = ProfileImageView()
  
  // Optional - For configuration purposes
  init() {
    profileImageView.properties.placeHolder = Image(systemName: "person.circle.fill")
    profileImageView.imageTapped = { print("Image tapped") }
  }
  
    var body: some View {
      profileImageView
        .frame(width: 100, height: 100)
    }
}
```

### Example: Implementation in UIKit
```
  var profileImageView = ProfileImageView()
  private func addProfileImageView() {
    let hostingController = UIHostingController(rootView: profileImageView)
    
    addChild(hostingController)
    view.addSubview(hostingController.view)
    
    hostingController.view.translatesAutoresizingMaskIntoConstraints = false
    hostingController.view.backgroundColor = .clear
    let constraints = [
      hostingController.view.centerXAnchor.constraint(equalTo: view.centerXAnchor),
      hostingController.view.centerYAnchor.constraint(equalTo: view.centerYAnchor),
      hostingController.view.widthAnchor.constraint(equalToConstant: 100),
      hostingController.view.heightAnchor.constraint(equalToConstant: 100)
    ]
    
    NSLayoutConstraint.activate(constraints)
    hostingController.didMove(toParent: self)
    
    profileImageView.imageTapped = {
      print("Image tapped")
    }
    
    profileImageView.badgeTapped = {
      print("Badge tapped")
    }
    
    hostingController.rootView = profileImageView
  }
```

Accessing methods & properties of `ProfileImageView`
```
    profileImageView.set(URL(string: "URL String"))
    profileImageView.properties.badging = .some(badge: .init(image: Image(systemName: "plus"), backgroundColor: .green, borderColor: nil, borderWidth: nil))
```


Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+    case none
+    case some(badge: BadgeStyle)
+  }
+  
+  public enum BackgroundType {
+    case plain(Color)
+    case linearGradient(LinearGradient)
+    case radialGradient(RadialGradient)
+    case angularGradient(AngularGradient)
+  }
```
borut-t, line 47: Can we put this under `ProfileImageView` extension inside this file instead?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+        ProfileImage(properties: properties, profileTapped: triggerImageTapClosure)
+        Badge(style: properties.badging, size: .init(width: geo.size.width / 3.5, height: geo.size.height / 3.5), badgeTapped: tiggerBadgeTapClosure)
+      }
+    }
+  }
+  
+  private func triggerImageTapClosure() { imageTapped?() }
+  private func tiggerBadgeTapClosure() { badgeTapped?() }
+  
+  private struct Badge: View {
```
borut-t, line 63: Under extension? Also, rename to `BadgeView` 🤷‍♂️

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+        .clipShape(Circle())
+        .gesture(TapGesture().onEnded({ badgeTapped() }))
+        .overlay(Circle().stroke(
+          (style.borderColor != nil) ? style.borderColor! : Color.white,
+          lineWidth: (style.borderWidth != nil) ? style.borderWidth! : 2)
+        )
+    }
+  }
+  
+  private struct ProfileImage: View {
```
borut-t, line 95: Under extension? And rename to `ImageView` 🤔 Thoughts?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+// MARK: - Public Methods
+@available(iOS 13, *)
+public extension ProfileImageView {
+  func set(_ url: URL?) {
+    switch url {
+    case .some(let url):
+      URLSession.shared.dataTask(with: url) { data, response, error in
```
borut-t, line 147: I like what @tonikocjan did. Can you check [it](https://github.com/poviolabs/PovioKit/blob/develop/Sources/Core/Utilities/ImageSource/ImageSource.swift#L37) out and see if we can use it here as well?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+//
+//  Created by Arlind Dushi on 3/9/22.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 13, *)
+public class ProfileImageProperties: ObservableObject {
+  @Published public var placeHolder = Image(systemName: "person")
```
borut-t, line 13: Typo - `placeHolder` vs `placeholder`

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+    let borderColor: Color?
+    let borderWidth: CGFloat?
+  }
+  
+  public enum BadgingMode {
+    case none
+    case some(badge: BadgeStyle)
+  }
+  
+  public enum BackgroundType {
```
borut-t, line 42: Can we just name it `Background`?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+public class ProfileImageProperties: ObservableObject {
+  @Published public var placeHolder = Image(systemName: "person")
+  @Published public var backgroundType = ProfileImageView.BackgroundType.plain(.clear)
+  @Published public var cornerRadius: CGFloat = 0
+  @Published public var contentMode: ContentMode = .fit
+  @Published public var borderColor: Color = .clear
+  @Published public var borderWidth: CGFloat = 0
+  @Published public var badging: ProfileImageView.BadgingMode = .none
+  @Published public var badgeAlignment: Alignment = .bottomTrailing
+  @Published var urlData: Data?
```
borut-t, line 21: Why did we go with `Data` type here?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+@available(iOS 13, *)
+public class ProfileImageProperties: ObservableObject {
+  @Published public var placeHolder = Image(systemName: "person")
+  @Published public var backgroundType = ProfileImageView.BackgroundType.plain(.clear)
+  @Published public var cornerRadius: CGFloat = 0
+  @Published public var contentMode: ContentMode = .fit
+  @Published public var borderColor: Color = .clear
+  @Published public var borderWidth: CGFloat = 0
+  @Published public var badging: ProfileImageView.BadgingMode = .none
+  @Published public var badgeAlignment: Alignment = .bottomTrailing
```
borut-t, line 20: Interesting approach. I would prefer going with one reference e.g. `Badge` where you can define all values, alignment as well 🤷‍♂️ 
What do you think?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+  @Published var urlData: Data?
+}
+
+@available(iOS 13, *)
+public struct ProfileImageView: View {
+  public var imageTapped: (() -> Void)?
+  public var badgeTapped: (() -> Void)?
+  @ObservedObject public var properties = ProfileImageProperties()
+  
+  public struct BadgeStyle {
```
borut-t, line 30: Rename to just `Badge` 🤷‍♂️ Thoughts?

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,168 @@
...
+    let image: Image
+    let backgroundColor: Color
+    let borderColor: Color?
+    let borderWidth: CGFloat?
+  }
+  
+  public enum BadgingMode {
```
borut-t, line 37: Rename to `Mode` and put it inside `Badge` 🤷‍♂️ Thoughts?

borut-t (commented):
Great work. I think we should create a really solid component, that's why so many comments/suggestions. Let's try to clear them out.

Resources/Views/ProfileImageView/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Example: Implementation in SwiftUI
+```
+struct ContentView: View {
+  var profileImageView = ProfileImageView()
+  
+  // Optional - For configuration purposes
+  init() {
+    profileImageView.properties.placeHolder = Image(systemName: "person.circle.fill")
+    profileImageView.imageTapped = { print("Image tapped") }
```
borut-t, line 13: I few more comments about the behavior:
- `profileView.properties.backgroundType = .plain(.red)` doesn't properly set the background. Background should be only inside the circle
Screenshot 2022-03-21 at 08 30 17

- border is set incorrectly. It should be set on the circle
Screenshot 2022-03-21 at 08 32 20

- if I do not set `placeHolder`, there is some a default picture in there
Screenshot 2022-03-21 at 08 34 17

- cornerRadius should be an enum defining perfect "circle" or custom value, so if i want a circle i don't want to handle the radius all the time when the view resizes

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
tonikocjan: Maybe we can use `Delegated`, which automatically handles reference cycles, instead of plain closures for callbacks.

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
ArlindDushi: It looks like that requires the app to have `kingfisher` no? If that framework isn't present should we fallback to this implementation then? What are your thoughts?

borut-t (approved):
I really like the changes. And pretty much everything works now 👌

borut-t:
Another minor thing. When running in UIKit env, there is a white background around it.
IMG_1431

Here is the code
```swift
profileView.properties.borderColor = .blue
profileView.properties.borderWidth = 2
profileView.properties.urlData = try? Data(contentsOf: URL(string: "https://misterstocks.com/wp-content/uploads/2019/07/00-tout-gaga.jpg")!)
profileView.properties.badging = .some(badge: .init(image: Image(systemName: "person.circle"), backgroundColor: .red, borderColor: .black, borderWidth: 1, alignment: .bottomTrailing))
```

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
tonikocjan: It does not require it. It just provides a default implementation of the `resolve` method when Kingfisher is available. You can provide a default implementation when it's not available.

borut-t: Actually I like this approach. Usually we use Kingfisher on every project where images are involved. So that is the happy path. But if for some reason there is no Kingfisher dependency, we can still provide our custom implementation.

tonikocjan: I would extend `ImageSource` and add `resolve` method implemented using `URLSession.shared.dataTask(with: url)` like it's done now.

borut-t: Great. @ArlindDushi can we do that?

ArlindDushi: @borut-t  Yes, I like the idea. Will work on that.

Sources/Core/Utilities/ImageSource/UrlImage.swift
```diff
@@ -0,0 +1,77 @@
+//
+//  UrlImage.swift
```
borut-t, line 2: Let's sync this name with the name of the class.

borut-t (approved):
I like it 👌

ArlindDushi:
> Another minor thing. When running in UIKit env, there is a white background around it. IMG_1431
> 
> Here is the code
> 
> ```swift
> profileView.properties.borderColor = .blue
> profileView.properties.borderWidth = 2
> profileView.properties.urlData = try? Data(contentsOf: URL(string: "https://misterstocks.com/wp-content/uploads/2019/07/00-tout-gaga.jpg")!)
> profileView.properties.badging = .some(badge: .init(image: Image(systemName: "person.circle"), backgroundColor: .red, borderColor: .black, borderWidth: 1, alignment: .bottomTrailing))
> ```

Looks like when `UIHostingController` is used we need to make its view background clear. I'll update documentation to include this part.

Sources/Core/Utilities/ImageSource/UrlImage.swift
```diff
@@ -16,6 +16,7 @@ class UrlImageModel: ObservableObject {
   
   init(url: URL?) {
     self.url = url
+    // TODO: Load from cache
```
borut-t: I would not try to replicate the whole Kingfisher logic. Let's just do without. So if dev decides to go without Kingfisher, he'll only have bare bone image loading mechanism.

borut-t (approved):

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -22,8 +22,8 @@ public class ProfileImageProperties: ObservableObject {
 
 @available(iOS 14, *)
 public struct ProfileImageView: View {
-  public var imageTapped = Delegated<Void, Void>()
-  public var badgeTapped = Delegated<Void, Void>()
```
borut-t: What happened to `Delegated`?

ArlindDushi: I reverted to closures b/c `Delegated` pattern doesn't work with `struct` (SwiftUI Views are structs).

tonikocjan: Good point; value types do not cause reference cycles so regular closures should be okay.

borut-t:
@ArlindDushi can we wrap this up?

ArlindDushi:
@borut-t yes, there's no more work left to be done here.

borut-t:
Great. Let's ping @poviolabs/ios-developers for some more reviews.

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,195 @@
...
+  private func triggerImageTapClosure() { imageTapped?() }
+  private func tiggerBadgeTapClosure() { badgeTapped?() }
+}
+
+@available(iOS 14, *)
+private extension ProfileImageView {
+  struct ImageView: View {
+    @ObservedObject var properties: ProfileImageProperties
+    var profileTapped: (() -> Void)
+    var optionalPlaceHolder: UIImage?
```
markom-povio, line 49: This is not used
Probably leftover :)

yllfejziu (approved):

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,195 @@
...
+    func getCornerRadius(for cornerType: ProfileImageView.CornerRadiusType) -> CGFloat {
+       switch cornerType {
+       case .circle:
+         return .infinity
+       case .custom(let cornerRadius):
+         return cornerRadius
+       }
+     }
+    
+    func getShape(for size: CGSize) -> AnyView {
```
markom-povio, line 94: Maybe we should avoid using _AnyView_ if not needed (see here for reference: https://www.swiftbysundell.com/articles/avoiding-anyview-in-swiftui/)

For example, here we can use something like this:
```
@ViewBuilder
    var imageBackground: some View {
      switch properties.backgroundType {
      case .plain(let backgroundColor):
        backgroundColor
      case let .linearGradient(gradient):
        gradient
      case let .angularGradient(gradient):
        gradient
      case let .radialGradient(gradient):
        gradient
      }
    }
```

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
```diff
@@ -0,0 +1,195 @@
...
+    
+    func constructProfileImage(placeholder: Image?) -> Image {
+      if let unwrappedPlaceholder = placeholder {
+        return unwrappedPlaceholder
+      }
+      
+      return Image(uiImage: UIImage())
+    }
+    
+    func getBackground() -> AnyView {
```
markom-povio, line 72: Maybe we should avoid using _AnyView_ if not needed (see here for reference: https://www.swiftbysundell.com/articles/avoiding-anyview-in-swiftui/)

For example, here we can use something like this:
```
@ViewBuilder
    var imageBackground: some View {
      switch properties.backgroundType {
      case .plain(let backgroundColor):
        backgroundColor
      case let .linearGradient(gradient):
        gradient
      case let .angularGradient(gradient):
        gradient
      case let .radialGradient(gradient):
        gradient
      }
    }
```

markom-povio (approved):
Nice 👌 
Some minor things to think about. Other than that, good work!

Sources/Core/Views/ProfileImageView/ProfileImageView.swift
ArlindDushi: Great suggestion! Thanks 🙌

PR #153, opened 2022-04-07
Feature/Money

markom-povio:
A swift implementation of the popular js library [Dinero.js](https://dinerojs.com/)
~~TODO~~:
- [x] Add Unit tests
- [x] Add README
- [x] Add examples
---
### Getters:
- cents: Cents
- currency: Currency
- localeIdentifier: String
- precision: Int
- unitValue: Double
- formatted: String?
- locale: Locale
### Manipulation:
- +, -, *, /
- multiply()
- divide()
- percentage()
- setLocaleIdentifier()
### Muttating Manipulations:
- multiply()
- divide()
- percentage()
- setLocaleIdentifier()
### Testing
- ==, >, >=, <, <= 
- isPositive
- isNegative
- isZero
- hasSameAmount()
- hasSameCurrency()
---
Additional: Confirmation to the protocols: Equatable, Comparable, Hashable, Codable, ExpressibleByIntegerLiteral, CustomStringConvertible


Sources/Core/Utilities/Money/Money.swift
```diff
@@ -0,0 +1,433 @@
...
+  func add(_ addend: Money) -> Money? {
+    guard addend.getCurrency() == getCurrency() else {
+      return .none
+    }
+    let normalizedPrecision = Money.normalizePrecision([self, addend])
+    let newAmount = normalizedPrecision[0].getAmount() + normalizedPrecision[1].getAmount()
+    return Money(amount: newAmount,
+                 currency: getCurrency(),
+                 localeIdentifier: getLocaleIdentifier(),
+                 precision: normalizedPrecision[0].getPrecision())
```
tonikocjan, line 112: You can access properties directly instead of via getters.

tonikocjan (changes requested):
Good work on this @markom-povio.

I think we can simplify the calculations a bit. If you consider that every `Money` value can be converted to _cents_ ( `precision` == 2 => `amount` == `cents`), regardless of the precision, we can implement the calculations in a very straightforward manner:

```swift

// pseudo-implementation

typealias Cents = UInt

extension Money {
  var cents: Cents {
    Cents(Double(amount) * pow(10, Double(-precision + 2)))
  }
}

  static func < (_ lhs: Self, _ rhs: Self) -> Bool {
    lhs.cents < rhs.cents
  }
  
  static func == (_ lhs: Self, _ rhs: Self) -> Bool {
    lhs.cents == rhs.cents
  }
  
  static func + (_ lhs: Self, _ rhs: Self) -> Self {
    precondition(lhs.currency == rhs.currency)
    return .init(amount: lhs.cents + rhs.cents, currency: lhs.currency, precision: .default)
  }
  
  static func - (_ lhs: Self, _ rhs: Self) -> Self {
    precondition(lhs.currency == rhs.currency)
    return .init(amount: lhs.cents - rhs.cents, currency: lhs.currency, precision: .default)
  }
  
  static func * (_ lhs: Self, _ rhs: Self) -> Self {
    precondition(lhs.currency == rhs.currency)
    return .init(amount: lhs.cents * rhs.cents, currency: lhs.currency, precision: .default)
  }
  
  static func / (_ lhs: Self, _ rhs: Self) -> Self {
    precondition(lhs.currency == rhs.currency)
    return .init(amount: lhs.cents / rhs.cents, currency: lhs.currency, precision: .default)
  }
```

I would also like to see conformance to the following protocols:
- Equatable
- Comparable
- Hashable
- Codable
- `ExpressibleByIntegerLiteral`
- `CustomStringConvertible`
- possibly `ExpressibleByStringLiteral` (some parsing logic would be required for this)

Furthermore, currency expressed as a `String` feels a bit error-prone. Let's think a bit how / if we can make it better. Maybe an enumeration of all supported currencies and we add new ones when needed.
Something like this maybe:

```swift
enum Currency: String, Equatable, Hashable {
  case dollar = "USD"
  case unknown // TODO: - handle currencies
  
  var unit: String {
    switch self {
    case .dollar:
      return "$"
    case .unknown:
      return ""
    }
  }
}
```

Sources/Core/Utilities/Money/Money.swift
```diff
@@ -0,0 +1,433 @@
...
+  func setLocaleIdentifier(_ locale: String) -> Money {
+    Money(amount: getAmount(),
+          currency: getCurrency(),
+          localeIdentifier: locale,
+          precision: getPrecision())
+  }
+}
+
+// MARK: - Public Methods - Muttating Manipulations
+public extension Money {
```
tonikocjan, line 200: I would remove this extension and make properties public instead.

markom-povio:
Thanks @tonikocjan for the review!
Amount is actually representing cents (or any other minor currency units).
One problem here is that Money is represented by the both amount (cents) and precision. For example, same amount value with different precision will get different Money value:
```
// This represents 1.0545$
Money(amount: 10545, currency: "USD", precision: 4).unitValue
// This represents 105.45$
Money(amount: 10545, currency: "USD", precision: 2).unitValue
```
This is why we need to "normalizePrecision" (to convert them to the same, maximum precision of the Money array without losing the value) and round it with appropriate rounding mode before any manipulation 
What we can do instead, is to use this `unitValue` for all manipulations
```
var unitValue: Double {
    amount.double / pow(10, precision.double)
}
```
I will check this and make appropriate changes ;)

PS I agree for the currency representation in String to be error prone, I will look to improve it

markom-povio:
I have introduced _Currency_ protocol with the _CurrencyCode_ enum to get more control over the initialisation
We can not init Money like this:
`Money.init(cents: 200, currencyCode: .eur)`

borut-t (commented):
I really like the solution overall. Nicely done and commented all parts. Also tests are there.
One thing that brought extra attention is currency enum. I've added a comment on that besides some other minor comments as well. I would like to hear your feedback about my comments.

Sources/Core/Utilities/Money/Money+Currency.swift
```diff
@@ -0,0 +1,112 @@
...
+  var symbol: String { get }
+}
+
+public extension Currency {
+  var currencyCode: String {
+    code.stringValue
+  }
+}
+
+public enum CurrencyCode: Codable {
```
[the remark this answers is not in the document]
markom-povio: @borut-t 
Actually yes, this is much cleaner 
My approach was complicated because I started implementing it without enum, I was passing String to the generator to get correct Currency. At first, I was thinking to leave it open for possible variations (like "en_US", "usd", "USD", etc), but that would be to complicated so I switched to the enum but did not notice that generator is not needed any more...
Thanks, I will change it

Sources/Core/Utilities/Money/Money.swift
```diff
@@ -8,31 +8,20 @@
...
-    case .currencyNotSame:
-      return "Currencies must be the same!"
-    }
-  }
-}
-
 public struct Money {
+  enum Constants {
+    static let defaultPrecision: Int = 2
+  }
```
borut-t, line 14: One small nit. Let's put it inside the extension next to Error enum.

borut-t (approved):
Great 🙌

MAKA02:
carding and loading is an act to be done by legitimate people 
need verification on how to get quick cc , PayPal transfers and more ?
have you been looking for a place to buy cc and get legit money TRANSFER services like PayPal,  western union and more ?
visit https://t.me/credit_card_uk

MAKA02:
carding and loading is an act to be done by legitimate people 
need verification on how to get quick cc , PayPal transfers and more ?
have you been looking for a place to buy cc and get legit money TRANSFER services like PayPal,  western union and more ?
visit https://t.me/credit_card_uk

borut-t:
@markom-povio, @tonikocjan have you managed to sit down and agree on the changes?

markom-povio:
@borut-t unfortunately no.
@tonikocjan had prepared some improvements and simplification but we did not get the chance to sync about it.

markom-povio:
Latest update:
- Added @tonikocjan improvements
- Code cleanup

nagavcindriqim (approved):

Sources/Core/Utilities/Money/Money.swift
```diff
@@ -0,0 +1,291 @@
...
+/// - Parameter rhs: An integer multiplier.
+public func * (_ lhs: Int, _ rhs: Money) -> Money {
+  var res = rhs
+  res.amount *= lhs
+  return res
+}
+
+///
+
+fileprivate func alignToSamePrecision(m1: inout Money, m2: inout Money) {
```
borut-t, line 283: Should here be a comment?

Sources/Core/Utilities/Money/Money.swift
```diff
@@ -0,0 +1,291 @@
...
+  
+  public var description: String {
+    return formatted ?? "\(amount) \(currency.symbol)"
+  }
+}
+
+// MARK: - Math operators
+
+public func + (_ lhs: Money, _ rhs: Money) -> Money {
+  precondition(lhs.currency == rhs.currency)
```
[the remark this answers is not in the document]
markom-povio: I would usually go with the comment, but the user will not see the comments for this method because XCode threat it as a basic operation (the "control + space" does not have any results) :/ 
On the other hand, the throw will add more complexity to the implementation, so I am not sure what is the best solution here.
Maybe to go with the precondition with the comment for now?

Sources/Core/Utilities/Money/Money.swift
markom-povio: Yes, I will add it

Sources/Core/Utilities/Money/Money.swift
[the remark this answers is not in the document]
markom-povio: Ok, lets go with the throw!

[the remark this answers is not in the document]
borut-t: There is no optimal solution here. Eitherway, the edge cases are crucial here. 

Going with `precondition` to avoid cumbersomeness, the app will crash.
Going with `throw` to avoid crashing, a cumbersome experience.

Let's discuss this in a pod meeting today.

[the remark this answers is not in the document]
tonikocjan: 1. App crash is not always _bad_. At least it indicates that you made a mistake and can fix it.
2. We shouldn't design libraries in a way that will make it painful for users to use just because some (non-diligent) programmers can use them in an incorrect way.
3. We can swap `precondition` with `assert` so that release builds remove the check. You are still able to catch bugs during development.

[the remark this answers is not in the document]
tonikocjan: However, I'm not confident that 3) is a good solution. If for some reason, in production, it just so happens that math is performed on two Money values with a different currency, the bug that will appear will be very subtle but the overall mistake might be much worse than a crash. For instance, if we display an incorrect amount of money to the user and they make an action based on that, that can have big consequences.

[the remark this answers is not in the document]
borut-t: I would go with `fatalError` here because I would like to see app termination in release builds as well.
```swift
guard lhs.currency == rhs.currency else { fatalError("When adding Money, both currencies must equal!" }
```

Resources/Core/Utilities/Money/README.md
```diff
@@ -0,0 +1,150 @@
...
+  /// Canadian Dollar (CAD)
+  case cad
+  /// Chinese Yuan Renminbi (CNY)
+  case cny
+  /// Japanese Yen (JPY)
+  case jpy
+  /// British Pound (GBP)
+  case gbp
+  /// Swiss Franc (CHF)
+  case chf
```
tonikocjan, line 65: Should we add an `unknown` case for other currencies?

tonikocjan (approved):
💯

Resources/Core/Utilities/Money/README.md
markom-povio: Not sure 🤔
What will be the currency code and symbol for that case?

borut-t: Maybe just leave it as is. We'll add new currencies once we get to a situation.

Sources/Core/Utilities/Money/Money.swift
```diff
@@ -280,47 +261,47 @@ public func * (_ lhs: Money, _ rhs: Money) throws -> Money {
...
 }
 
 /// A math operation to multiply the `Money` object with the Int multiplier.
 /// - Important: Both objects need to be of the same `currency`!
 /// - Parameter rhs: An integer multiplier.
-/// - Throws: `Money.Error.currencyNotSame` when the currencies are not the same
+/// - Important: *FatalError* will be raised when the currencies are not the same
 /// - Returns: **New** Money object
 public func * (_ lhs: Money, _ rhs: Int) -> Money {
   var res = lhs
```
borut-t, line 307: Hm, is this comment correct? 🤔

Sources/Core/Utilities/Money/Money.swift
```diff
@@ -0,0 +1,330 @@
...
+}
+
+/// A math operation to division of the cents and the `Money` object.
+/// - Important: Both objects need to be of the same `currency`!
+/// - Parameter rhs: The amount of cents to be divided. It is converted to `Money`:  Money(amount: lhs, currency: rhs.currency, precision: defaults.precision)
+/// - Returns: **New** Money object
+public func - (lhs: Money.Cents, rhs: Money) -> Money {
+  let lhs = Money(amount: lhs, currency: rhs.currency, localeIdentifier: rhs.localeIdentifier, precision: defaults.precision)
+  return lhs - rhs
+}
```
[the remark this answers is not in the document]
borut-t: Typo...
`The amount of cents to be divided.` -> `The amount of cents to be substracted.`

borut-t (approved):
Awesome work @markom-povio and @tonikocjan. I've commented on a few nits. After that, we can 🚢 it 💪

borut-t:
@markom-povio can you merge this ticket?

Issue #155, opened 2022-04-13
[UI] Dialogue


borut-t:
A UI component that effectively replaces and extends native UIActionSheet.

Here is the list of capabilities that it should cover:
* dynamically expandable based on the content height
* pan gesture to dismiss
* tap gesture behind it to dismiss
* custom transition animations / can be expandable
* custom presentations; it can be presented at the bottom like a regular action sheet or centered as a modal

PR #156, opened 2022-04-18
Fix: [UI] BottomSheet

markom-povio:
Improvements and PR fixes for [UI] BottomSheet
---
- Added separate file for WizardDataSource.swift
- Changed year in Copyright header
- Added private extension in WizardContentView
- Change name of the component from "Wizard" to the "BottomSheet"
- BottomSheet: Added outside tap gesture to dismiss (swipe to dismiss is already available with modal presentation)
- BottomSheetContentView: addHelperGestureRecognizer to skip "tapToDismiss" on content view
- Added README file
- Rename for BottomSheetAnimators file

> Note: Simple Demo project is on separate branch: [BottomSheet Demo](https://github.com/poviolabs/PovioKit/tree/chore/demo_bottom_sheet)


tonikocjan:
@markom-povio there is a lot of stuff here that I think should not be part of this PR? Can you check.

Resources/Controllers/BottomSheet/README.md
```diff
@@ -0,0 +1,159 @@
+#  BottomSheet
```
tonikocjan, line 1: Why the name change? [Wizard](https://docs.microsoft.com/en-us/windows/win32/uxguide/win-wizards) is a common name used for a multi-step user interface.

Resources/Controllers/BottomSheet/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+A UI component that effectively replaces and extends native UIActionSheet.
```
tonikocjan, line 3: I think `Wizard` is a bit of an overkill as a replacement for UIActionSheet as it prepares a lot of stuff regarding animations between different steps and so on.

Resources/Controllers/BottomSheet/README.md
```diff
@@ -0,0 +1,159 @@
+#  BottomSheet
+
+A UI component that effectively replaces and extends native UIActionSheet.  
+
+## Overview  
+
+BottomSheet has the following characteristics:
+- dynamically expandable based on the content height
+- swipe down gesture to dismiss
+- tap gesture behind the content to dismiss
```
tonikocjan, line 10: These (the gestures) should be left for the users to implement/integrate in any way they want. The Wizard itself should be as self-contained as possible. It can actually be used as an embedded view controller inside another to provide context for a wizard(multi step)-like UI.

Resources/Controllers/BottomSheet/README.md
```diff
@@ -0,0 +1,159 @@
...
+- **LazyStep**: 
+```swift
+typealias LazyStep = () -> BottomSheetStep
+```
+- **AnimatorFactory**: *Enum that holds predefined transition Animations*
+```swift
+enum AnimatorFactory
+```
+
+### Default animations:
```
tonikocjan, line 40: I would change wording from `Default` to `Provided` or `Predefined` or smth similar.

Resources/Controllers/BottomSheet/README.md
```diff
@@ -0,0 +1,159 @@
...
+- slideInSlideOut
+- fadeIn
+- fadeOut
+- fadeInFadeOut
+- popInPopOut  
+
+Besides these predefined animations, it is possible to combine two or more animations with the available `composite` function, but also to create custom animation with return type: ```typealias Animator = (UIView, UIView) -> Void```
```
tonikocjan, line 47: `Besides these predefined animations, it is possible ...` -> `Furthermore, it is possible ...`.

tonikocjan (commented):
Thanks for your work on this @markom-povio. Left a couple of comments to look at.

markom-povio:
@tonikocjan I will reply for most of the comments here because they are more or less related:

> Why the name change? [Wizard](https://docs.microsoft.com/en-us/windows/win32/uxguide/win-wizards) is a common name used for a multi-step user interface.  

I do not mind having the *Wizard* name, the idea was to align with the parent issue [BottomSheet](https://github.com/poviolabs/PovioKit/issues/155)

> I think Wizard is a bit of an overkill as a replacement for UIActionSheet as it prepares a lot of stuff regarding animations between different steps and so on.  

Yes, probably we do not need animations (or even the Steps and transition between them) if we want simple UIActionSheet replacement. But we have here a nice solution that covers both scenarios if we want to present simple or more complex bottomSheet UI.  

> These (the gestures) should be left for the users to implement/integrate in any way they want. The Wizard itself should be as self-contained as possible. It can actually be used as an embedded view controller inside another to provide context for a wizard(multi step)-like UI.  

The swipe gesture we already have out of the box with presenting modal BottomSheetViewController, but the tap gesture was one of the requirements in the parent issue [BottomSheet](https://github.com/poviolabs/PovioKit/issues/155)

Basically, most of the questions are on point here, but the solution is depending of the idea behind the component.  
Do we need this complexity of having the Steps, animations and tap gesture and should we rename the component back to the *Wizard* name so it better reflects the behaviour (I think this should be done if all the logic stays as it is)?

@borut-t what do you think?

borut-t:
Thanks @tonikocjan and @markom-povio for digging in. I think we've made a small mistake here since we've actually joined two components together where they should be separated. A wizard and a dialogue. I've discuss this with Toni this morning and we would like to create both instead. @markom-povio I'll update the ticket wording so it will better suit it's function as it is currently. Sorry for this, but I was a bit lame here not clearly check the existing implementation from Toni.

PR #160, opened 2022-04-27
UI/Dialog

markom-povio:
### Component requirements:
A UI component that effectively replaces and extends native UIActionSheet.   

- [x] dynamically expandable based on the content height
- [x] pan gesture to dismiss
- [x] tap gesture behind it to dismiss
- [x] custom transition animations / can be expandable
- [x] custom presentations; it can be presented at the bottom like a regular action sheet or centered as a modal
---
### TODO:
- [x] Demo project is created on separate branch: [Dialog demo](https://github.com/poviolabs/PovioKit/tree/chore/dialog_demo) 
- [x] Documentation
---
Demo video (updated 14.10.2022.):

https://user-images.githubusercontent.com/95407653/195816096-0b1271aa-7e2c-4734-b107-ae24a08388ca.MP4


Sources/UI/Dialog/DialogContentView.swift
```diff
@@ -0,0 +1,138 @@
...
+
+/// UIView that holds user-defined custom UI
+/// In subclass, use ``content`` UIView to addSubview
+///
+/// ```swift
+/// //Example:
+/// content.addSubview(button)
+/// ```
+open class DialogContentView: UIView {
+
```
borut-t, line 20: Redundant new line.

Sources/UI/Dialog/DialogTransitionAnimation.swift
```diff
@@ -0,0 +1,33 @@
...
+  case flip
+  case pushPop
+  case custom(DialogTransitionAnimation)
+}
+
+/// Protocol that defines Dailog animations.
+///
+/// Use it when creating custom transition animation.
+public protocol DialogTransitionAnimation: UIViewControllerAnimatedTransitioning {
+
```
borut-t, line 24: Redundant new line.

borut-t (commented):
Great work 🙌  Here are a few remarks I would like to see implemented before we wrap this up:
- be able to dismiss it by gesture
- be able to set the background color (dim) behind the dialog
- be able to change the width or horizontal inset of the dialog
- a readme file maybe explaining some basics how to create it

Also great work on animations. Since we are providing a mechanism to include custom animation, we don't need to include three extra animations by default. I would go with standard and maybe one more, like fade animation. All others are pretty explicit and will probably never be used in our projects.

The Demo project is also a great thing so any developer can check it out up-front and play around. I was also planning to have a demo for all our features inside, so this is a great step towards.

nagavcindriqim (commented):
Great work!

borut-t:
I've just tested this out and here is the report:
- i don't see a "swipe down" gesture to dismiss dialogue, and it should support interactive dismiss
- i would like to fixate the height of the dialogue, sometimes it half the screen height or any arbitrary height; the content should scroll inside dialogue
- when presenting the dialogue, there is some semitransparent background on the viewcontroller
- in the demo app, when use "simple animation" with "bottom" type, scrolling to the top will reveal the rootController at the bottom of the screen

markom-povio:
Added:
- "swipe down to dismiss" interactive gesture.
- option to set a custom height
- fix for *scrolling to the top will reveal the rootController at the bottom of the screen*

Info regarding this issue: 
> when presenting the dialogue, there is some semitransparent background on the viewController

This is set in the demo project with: `setDialogBackground()` but I have removed it to avoid confusion.

borut-t:
I appreciate your work on this @markom-povio, but we have moved our focus towards SwiftUI components. Since this can be easily done with [.sheet](https://developer.apple.com/documentation/swiftui/view/sheet(ispresented:ondismiss:content:)) modifier, I'll close this PR.

PR #161, opened 2022-05-20
feature/ParameterEncoder as parameter

EgzonArifi:
**Current issue:** 
- ANC request method explicitly sends Encodable object as query params on .get, .delete, .head http methods, but we want to easily change that from ANC usage

**Proposal:**
- In order to not break current usage, new method is created and we can pass ParameterEncoder without having to import anything else.

```client
      .request(method: .delete,
               endpoint: Endpoints.conversationParticipants(conversationId),
               encode: request,
               parameterEncoder: .jsonEncoder())
      .validate()
      .defaultFailure()
      .asVoid```

### DIFF
--- a/Sources/UI/Extensions/Image+Povio.swift
+++ b/Sources/UI/Extensions/Image+Povio.swift
@@ -14,6 +14,7 @@ public extension Image {}
 #if canImport(Kingfisher)
 import Kingfisher
 
+@available(iOS 14, *)
 private extension Image {
   func resolveWithKingFisher(from url: URL?, placeholder: Image?) -> some View {
     KFImage(url)
--- a/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
+++ b/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
@@ -111,6 +111,29 @@ public extension AlamofireNetworkClient {
     return .init(with: request, eventMonitors: eventMonitors)
   }
   
+  func request<E: Encodable>(
+    method: HTTPMethod,
+    endpoint: URLConvertible,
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: Sources/UI/Extensions/Image+Povio.swift [hunk 1]
```diff
@@ -14,6 +14,7 @@ public extension Image {}
 #if canImport(Kingfisher)
 import Kingfisher
 
+@available(iOS 14, *)
```
ArlindDushi, line 17: 😬

ArlindDushi (approved):
LGTM 👌

tonikocjan (approved):
Deprecate the current method, otherwise 👍

Sources/UI/Extensions/Image+Povio.swift
borut-t: Yeah, why that?

borut-t (approved):
👌

Sources/UI/Extensions/Image+Povio.swift
EgzonArifi: that was from an earlier PR and I'm supporting iOS 13+

borut-t: Can we change it?

EgzonArifi: you mean to support iOS 13+ for this extension?

ArlindDushi: Kingfisher is supported in SwiftUI from iOS14.0+  that's why we had to go with that min. version

borut-t: My bad. Thanks for explanation. All Good.

PR #162, opened 2022-05-31
OAuthInterceptor - Smaller memory footprint

tonikocjan:
Instead of storing `Request`, which is a reference type, therefore its memory never gets cleared for the entire lifetime of the app, store `UUID`, which is a value type (hence copied). This change will leave a smaller memory footprint when using `OAuthInterceptor`.


ArlindDushi (approved):
I like it 👌

PR #163, opened 2022-05-31
OAuthInterceptor - Use Semaphore instead of Lock

tonikocjan:
Replaced `NSLock` with `DispatchSemaphore`, which might help with some issues that are currently present when using `OAuthInterceptor`.

This is an experimental change that needs to be tested in-depth in the projects that use `OAuthInterceptor`.

### DIFF
--- a/Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
+++ b/Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
@@ -28,7 +28,7 @@ open class OAuthRequestInterceptor {
   private let provider: OAuthProvider
   private let storage: OAuthStorage
   private let adapter: OAuthHeadersAdapter
-  private let lock: NSLock = .init()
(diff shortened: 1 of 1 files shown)


Sources/Networking/AlamofireNetworkClient/OAuth/OAuthInterceptor.swift
```diff
@@ -28,7 +28,7 @@ open class OAuthRequestInterceptor {
   private let provider: OAuthProvider
   private let storage: OAuthStorage
   private let adapter: OAuthHeadersAdapter
-  private let lock: NSLock = .init()
+  private let lock: DispatchSemaphore = .init(value: 1)
```
yllfejziu, line 31: Maybe we should rename the variable name as well and change it from `lock` to  `semaphore`.

tonikocjan: I think it’s okay because we are using semaphore as a lock.

PR #165, opened 2022-06-03
ANC/Client-level error handling

tonikocjan:
Current issue with error handling is that you are able to specify error handlers per-request only. This PR makes it so that you can specify a default handler on a client level. The default handler can still be overridden if necessary.


EgzonArifi (changes requested):
Nice 👌, @tonikocjan can we also pass ErrorHandler here:
https://github.com/poviolabs/PovioKit/blob/1d289f011011eadb92893f3b0ab61bc96675c6c4/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift#L139

PR #170, opened 2022-06-23
Feature/PromiseKit - Sleep

borut-t:
I got an idea to have a delay after certain promise finishes up without interrupting the flow. So I write up and function that does this for you.

```swift
worker
  .job
  .sleep(duration: .milliseconds(500))
  .finally { ... }
```

Where is this helpful? I got a scenario where pull-to-refresh refreshed "to quickly". Yes, you heard it right. The API that was triggered on pull to refresh returned response before the component was rendered on screen. So adding a delay solved that issue so component had time to render on screen completely.

You would probably argue why handle UI stuff on the business part of the app. I completely feel you and I thought the same way. But after thinking about this issue way too long, I've decided to do it this way. If I would handle this on the UI, I would probably do smth like that:

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
  // update UI
}
```

But that's just silly. And doing that for every such task is silly as well. Instead, just delay API response a bit and you are done.

Thoughts?

### DIFF
--- a/Sources/PromiseKit/Promise.swift
+++ b/Sources/PromiseKit/Promise.swift
@@ -107,6 +107,7 @@ public class Promise<Value>: Future<Value, Error> {
   }
 }
 
+// MARK: - States
 public extension Promise {
   var isResolved: Bool {
     result != nil
@@ -153,6 +154,7 @@ public extension Promise {
   }
 }
 
+// MARK: - Utils
 public extension Promise {
   /// Convert this Promise to a new Promise where `Value` == ()
   var asVoid: Promise<()> {
@@ -199,6 +201,7 @@ public extension Promise {
   }
 }
 
+// MARK: - Core
 public extension Promise {
   /// Returns a composition of this Promise with the result of calling `transform`.
   ///
--- a/Sources/PromiseKit/Promise.swift
+++ b/Sources/PromiseKit/Promise.swift
@@ -189,7 +189,7 @@ public extension Promise {
         switch $0 {
         case .success(let value):
           dispatchQueue.asyncAfter(deadline: .now() + duration) {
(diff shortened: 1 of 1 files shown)


gentique:
I think the timer should start once the job starts executing, otherwise we risk delaying the UI too much (In the case where the fast API response decides to be slow).

ArlindDushi (approved):
Had once a similar problem where I had to delay the response. Nice to have this as part of the PovioKit 👌

borut-t:
> I think the timer should start once the job starts executing, otherwise we risk delaying the UI too much (In the case where the fast API response decides to be slow).

Maybe given example is not the best for this use case, but delay could server for other purposes as well.

Sources/PromiseKit/Promise.swift
```diff
@@ -176,8 +178,30 @@ public extension Promise {
...
+  func sleep(
+    duration: DispatchTimeInterval,
+    on dispatchQueue: DispatchQueue = .main
+  ) -> Promise<Value> {
+    Promise { seal in
+      self.finally {
+        switch $0 {
+        case .success(let value):
+          dispatchQueue.asyncAfter(deadline: .now() + duration) {
+            seal.resolve(with: value)
```
tonikocjan, line 192: Resolve the promise on the provided queue.

tonikocjan (approved):
I think this can be useful in certain scenarios.

Sources/PromiseKit/Promise.swift
borut-t: https://github.com/poviolabs/PovioKit/pull/170/commits/6bbec698589d7d97ae735f331dbbfaa881698a1f

PR #176, opened 2022-07-14
Feature/ci workflow improvement

gentique:
Recently I had to do some research about CI-related tasks on my project and I found out about some good practices.

I've split the `Generate Project` step into two:

- `build-without-testing`
- `test-without-building`

Now we have one step for building the package and one step for testing the package. This way we can accurately catch if it's failing during build or test time.

I've added  `set -o pipefail` which makes the pipeline use an occurred [error code as the final code](http://redsymbol.net/articles/unofficial-bash-strict-mode/) of the whole pipeline.

One other interesting argument is `xcpretty`. This formats the code keeps it clean thus we can be more focused on what we actually want from the logs.

### Left we have tests with xcpretty, on the right without.
Screen Shot 2022-07-09 at 10 47 48 AM

There was a caveat with this approach, since it could happen that xcpretty didn't print out the logs, a fix for this was setting `env NSUnbufferedIO=YES` , basically this allowed for timely feedback. You can find more information  [here](https://nshipster.com/launch-arguments-and-environment-variables/).

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -21,7 +21,7 @@ jobs:
         xcode-version: latest-stable
 
     - name: Build project
-      run: set -o pipefail xcodebuild build-for-testing -destination 'name=iPhone 13 Pro' -scheme 'PovioKit-Package'
+      run: set -o pipefail xcodebuild build-for-testing -destination 'name=iPhone 13 Pro' -scheme 'PovioKit-Package' | xcpretty
 
     - name: Run tests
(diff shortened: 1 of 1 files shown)


borut-t (approved):
Nice. This is way better. But don't we need to generate xcodeproj anymore?

gentique:
@borut-t I forgot to mention it in the PR description but `swift package generate-xcodeproj` is deprecated and you can open a swift package directly without the need to have a `.xcodeproj`.

PR #171, opened 2022-06-23
Feature/StoreKit extension

borut-t:
A shortcut to request a review popup from the current scene.

PR #177, opened 2022-07-18
Feature/storekit tests

gentique:
This PR is based on this comment https://github.com/poviolabs/PovioKit/pull/171#discussion_r908511815 

Though when I started developing, I noticed that you aren't able to create a `UIScene` by yourself, so I had to do some workarounds to get the desired behavior.

If you check the commits, you can see that I did some refactoring since I didn't like how some things turned out. I still think this can be done better, so I am very open to any suggestions!

### DIFF
--- a/Sources/Core/Extensions/Other/SKStoreReviewController+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/SKStoreReviewController+PovioKit.swift
@@ -60,11 +60,6 @@ public extension SKStoreReviewController {
 }
 
 @available(iOS 14.0, *)
-public protocol RequestReviewProviding {
-  func requestReview(in scene: Scene?)
-}
-
-@available(iOS 14.0, *)
 public protocol SceneProviding {
   func getConnectedScenes() -> [Scene]
 }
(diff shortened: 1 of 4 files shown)


ArlindDushi (approved):
LGTM 👌

borut-t:
@gentique I really like your enthusiasm to cover this with unit tests. But from experience and from this example you can clearly see we need to put too much effort in order to test native functionalities. I would skip this tbh.

gentique:
> @gentique I really like your enthusiasm to cover this with unit tests. But from experience and from this example you can clearly see we need to put too much effort in order to test native functionalities. I would skip this tbh.

It served me very well as an exercise since I haven't had a case where I needed to spy a type that you can't directly instantiate, therefore I'm perfectly fine with declining this PR.

gentique:
@borut-t On a second note, it is not the native functionality I'm testing per se, rather this piece of logic `first { $0.activationState == .foregroundActive } as? UIWindowScene` , since I can't assert if SKStoreController presents the reviewRequest popup, nor if UIApplication gives us any scenes back.

borut-t:
@gentique nice. We could save it maybe in the examples repo so we don't loose what you've worked on 🤷‍♂️

PR #179, opened 2022-08-08
Promise/Sequence

tonikocjan:
Implemented a `sequence` operator which executes tasks one by one. One could instead use `concurrentDispatch` with `concurrent` set to 1, but that involves unnecessary overhead of synchronization.

One example where this might be useful is, for instance, when you have to download several images and save them to Photos library, and the order is important.

### DIFF
--- a/Sources/PromiseKit/ConcurrentDispatch.swift
+++ b/Sources/PromiseKit/ConcurrentDispatch.swift
@@ -43,7 +43,7 @@ public func concurrentlyDispatch<T>(
   retryCount: Int = 2,
   on dispatchQueue: DispatchQueue? = .main
 ) -> Promise<()> {
-  return .init { seal in
+  .init { seal in
     let barrier = DispatchQueue(label: "com.poviokit.promisekit.barrier", attributes: .concurrent)
     
     var segmentIndex = concurrent
--- a/Sources/PromiseKit/ConcurrentDispatch.swift
+++ b/Sources/PromiseKit/ConcurrentDispatch.swift
@@ -32,6 +32,9 @@ import Foundation
 /// )
 /// .finally { print("Upload result: \($0)") }
 ///
+/// In scenarious where only one concurrent task can be executed at once,
+/// using `sequence` is preferable as it is much faster.
+///
 /// - Parameter next: Spawn a task with the given index. Return `nil` if all tasks have been spawn.
 /// - Parameter concurrent: The number of concurrent tasks executing at a given time.
 /// - Parameter retryCount: The number of times a task should be retried in case it fails.
--- a/Sources/PromiseKit/ConcurrentDispatch.swift
+++ b/Sources/PromiseKit/ConcurrentDispatch.swift
@@ -32,7 +32,7 @@ import Foundation
 /// )
 /// .finally { print("Upload result: \($0)") }
 ///
-/// In scenarious where only one concurrent task can be executed at once,
+/// In scenarios where only one concurrent task can be executed at once,
 /// using `sequence` is preferable as it is much faster.
 ///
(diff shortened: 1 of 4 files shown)


markom-povio (approved):
Looking good! 👍
Just a minor thing, we are missing copyright in the header.

Sources/PromiseKit/ConcurrentDispatch.swift
```diff
@@ -32,6 +32,9 @@ import Foundation
 /// )
 /// .finally { print("Upload result: \($0)") }
 ///
+/// In scenarious where only one concurrent task can be executed at once,
```
gentique, line 35: Typo `scenarious` ->`scenarios `

Tests/Tests/PromiseKit/PromiseTests.swift
```diff
@@ -1006,6 +1006,78 @@ extension PromiseTests {
   }
 }
 
+// MARK: - Sequence tests
+
+extension PromiseTests {
+  func testSequence1() {
+    func next(_ idx: Int) -> Promise<Int>? {
+      guard idx < 10 else { return nil }
+      return async(idx, delay: .random(in: 0.01...0.5))
```
gentique, line 1015: Any particular reason why this is 10? Also, that in combination with the potentially long random delay 0.5, causes this test to go longer than 2 seconds.

Tests/Tests/PromiseKit/PromiseTests.swift
```diff
@@ -1006,6 +1006,78 @@ extension PromiseTests {
...
+      { async(0, delay: .random(in: 0.01...0.5)) },
+      { async(1, delay: .random(in: 0.01...0.5)) },
+      { async(2, delay: .random(in: 0.01...0.5)) },
+      { async(3, delay: .random(in: 0.01...0.5)) },
+      { async(4, delay: .random(in: 0.01...0.5)) },
+    ]
+    let ex = expectation(description: "")
+    sequence(promises: promises)
+      .then { ex.fulfill() }
+    waitForExpectations(timeout: 10)
```
gentique, line 1063: I don't understand the value these tests provide given that the objective of `sequence` is to execute the work serially.

Tests/Tests/PromiseKit/PromiseTests.swift
```diff
@@ -1006,6 +1006,78 @@ extension PromiseTests {
...
+      { async(2, delay: .random(in: 0.01...0.5)) },
+      { async(NSError.err, Int.self, delay: .random(in: 0.01...0.5)) },
+      { async(4, delay: .random(in: 0.01...0.5)) },
+    ]
+    let ex = expectation(description: "")
+    sequence(promises: promises)
+      .catch { _ in ex.fulfill() }
+    waitForExpectations(timeout: 10)
+  }
+}
```
gentique, line 1079: I think the names for these tests are not up to standard/convention. Without reading what's actually being tested you can't know externally what's going on. In this case, it would be a testSequenceCatchesErrorOnGivenArrayWIthError or something similar.

gentique (changes requested):

Tests/Tests/PromiseKit/PromiseTests.swift
tonikocjan: Altered the interval to `0.01...0.05` so now the test runs for at most 0.5s.

borut-t (approved):
👌

Tests/Tests/PromiseKit/PromiseTests.swift
tonikocjan: https://github.com/poviolabs/PovioKit/pull/179/commits/e864a6254f7d75419da1e44ca85237d90d2a53d8

gentique (approved):

PR #180, opened 2022-08-09
PromiseKit/Replace synchronization primitive

tonikocjan:
### Replace `DispatchQueue` with `NSLock` as a locking primitive

Based on measurements, `NSLock` is 3x faster, and in the best case, even as much as 10x quicker than `DispatchQueue` using `sync(barrier:)` (`async(barrier:)` is much slower still).

---

## Benchmark 1:

This benchmark spawns different amount of threads, where each performs an operation (either read or write) multiple times. We see linear scalability with both locking primitives; however, `NSLock` scales much better (being around 3x faster).

read
Write

## Benchmark 2:

In this benchmark, we set up a more realistic test scenario that mimics usage in our apps. We performed three tests, a read-intensive test, a write-intensive test, and a mixed test. Following are the results:

### Read-intensive test:

**DispatchQueue:**
Avg: 0.41s
Std: 0.19s

**NSLock**:
Avg: 0.11s
Std: 0.07s

__A ~3.72x improvement.__

### Write-intensive test:

**DispatchQueue**:
Avg: 0.50s
Std: 0.03s

**NSLock**:
Avg: 0.20s
Std: 0.08s

__A ~2.5x improvement.__

### Mixed test:

**DispatchQueue**:
Avg: 4.27s
Std: 2.99s

**NSLock**:
Avg: 0.45s
Std: 0.12s

__A ~9,5x improvement.__

We analyzed the performance of `Mutex` and `RWLock` (read-write lock) too, but they achieve similar performance as `NSLock`, hence `NSLock` was chosen.

---

Additional testing on actual projects is welcomed. :)


gentique (approved):
Nice work 👏

Sources/PromiseKit/Future.swift
```diff
@@ -76,3 +88,21 @@ private extension Future {
...
+  }
+  
+  @inlinable
+  @inline(__always)
+  func write(_ work: () -> Void) {
+    lock()
+    defer { unlock() }
+    return work()
+  }
+}
```
borut-t, line 108: Can we move this under PromiseKit/Extensions folder instead?

Sources/PromiseKit/Promise.swift
```diff
@@ -873,6 +873,8 @@ public extension Promise where Value == Void {
 }
 
 extension Optional where Wrapped == DispatchQueue {
+  @inlinable
+  @inline(__always)
```
borut-t, line 877: Why using `inlinable` and `inline`?

borut-t (commented):
Great work. Have some second thoughts about `inline` attribute. Can you elaborate its usage?

Sources/PromiseKit/Promise.swift
tonikocjan: The way swift has this implemented is a bit confusing. The in-depth discussion can be seen here: https://forums.swift.org/t/when-should-both-inlinable-and-inline-always-be-used/37375/2

Sources/PromiseKit/Future.swift
tonikocjan: It's not part of the public API, and I think it should not be.

Sources/PromiseKit/Promise.swift
borut-t: I've read about it [here](https://swiftrocks.com/understanding-inlinable-in-swift.html) and [here](https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20170227/004886.html) as well. They don't recommend it. It's public API but undocumented.

Sources/PromiseKit/Future.swift
borut-t: What do you mean? It is already, ust put it in a separate file 🤔

Sources/PromiseKit/Promise.swift
tonikocjan: Removed @inlinable.

Sources/PromiseKit/Future.swift
tonikocjan: It is not. For it it be public it has to have a `public modifier`.

borut-t: But if it's internal access, other methods can still see it without exposing the interface right?

markom-povio (approved):

Sources/PromiseKit/Future.swift
tonikocjan: Decided to mark the extension as `fileprivate` to prevent others using it.

borut-t (approved):
Great work 🙌

PR #184, opened 2022-08-24
PromiseKit/Fix #183

### LINKED ISSUE #183: PromiseKit  - tests failing
I tried running the tests repeatedly and couldn't get them to fail, none the less, I suggest that we add the delay only for `testAllResolvedOnMainThread()` as an initial workaround until this is fully resolved. If the other tests perform as expected we shouldn't burden them with the additional delay.

``` swift
  func testAllResolvedOnMainThread() {
    let ex1 = expectation(description: "")
    let ex2 = expectation(description: "")
    all(on: .main, promises: [async(0, delay: 0.05 ,on: .main), async(1, delay: 0.05, on: .main)])
      .then { _ in
        XCTAssertTrue(Thread.isMainThread)
        ex1.fulfill()
      }
    all(on: .main, promises: [async(0, on: .main), async(NSError.err, Int.self, delay: 0.05, on: .main)])
      .catch { _ in
        XCTAssertTrue(Thread.isMainThread)
        ex2.fulfill()
      }
    waitForExpectations(timeout: 2)
  }
```

I did some further evaluation and for example `testAllListAsyncFailing` if we set the delay to 0. `func all<T, C>(on dispatchQueue: DispatchQueue? = .main, promises: C)` fails to stop on the first error promise. I don't want to hijack this PR with these discussions but I believe this needs evaluation since it's not performing as intended.

_Originally posted by @gentique in https://github.com/poviolabs/PovioKit/issues/139#issuecomment-1223620801_

tonikocjan:
Fixes: #183

### DIFF
--- a/Sources/PromiseKit/Future.swift
+++ b/Sources/PromiseKit/Future.swift
@@ -15,7 +15,7 @@ public class Future<Value, Error: Swift.Error> {
   private var internalResult: FutureResult?
 }
 
-public extension Future {
+internal extension Future {
   var result: FutureResult? {
     read {
       internalResult
@@ -24,10 +24,11 @@ public extension Future {
   
   func setResult(_ result: FutureResult?, on dispatchQueue: DispatchQueue? = nil) {
     write {
+      guard self.internalResult == nil else { return }
       self.internalResult = result
       guard self.isEnabled, let result = result else { return }
-      dispatchQueue.async {
-        self.observers.forEach { $0.notifity(result) }
+      for observer in self.observers {
+        dispatchQueue.async { observer.notifity(result) }
       }
     }
   }
--- a/Sources/PromiseKit/Promise.swift
+++ b/Sources/PromiseKit/Promise.swift
@@ -56,12 +56,10 @@ public class Promise<Value>: Future<Value, Error> {
   }
   
   public func resolve(with value: Value, on dispatchQueue: DispatchQueue? = .main) {
-    guard isAwaiting else { return }
     setResult(.success(value), on: dispatchQueue)
   }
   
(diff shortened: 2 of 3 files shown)


Sources/PromiseKit/Future.swift
```diff
@@ -24,10 +24,11 @@ public extension Future {
   
   func setResult(_ result: FutureResult?, on dispatchQueue: DispatchQueue? = nil) {
     write {
+      guard self.internalResult == nil else { return }
       self.internalResult = result
       guard self.isEnabled, let result = result else { return }
-      dispatchQueue.async {
-        self.observers.forEach { $0.notifity(result) }
+      for observer in self.observers {
+        dispatchQueue.async { observer.notifity(result) }
```
tonikocjan, line 31: The root of the problem is here. By keeping the loop on the thread that resolved the promise and only dispatching the observers themselves, tests work as expected.

Sources/PromiseKit/Promise.swift
```diff
@@ -56,12 +56,10 @@ public class Promise<Value>: Future<Value, Error> {
   }
   
   public func resolve(with value: Value, on dispatchQueue: DispatchQueue? = .main) {
-    guard isAwaiting else { return }
```
tonikocjan, line 59: Moving this check into the `Future.setResult(_,dispatchQueue:)` serves two purposes:
 - No need to acquire the lock twice, increasing the performance.
 - Potentially fixes a race condition, where a timely interrupt might suspend the thread right after `isAwaiting` returns, entering the critical section inside `setResult` twice.

gentique (approved):

borut-t (approved):

borut-t:
@tonikocjan can we merge this?

PR #185, opened 2022-08-31
UI/GradientView - additional initializer

tonikocjan:
It can be useful to provide an already configured `CAGradientLayer` during initialization.


borut-t (approved):
Great, like it 👌

markom-povio (approved):

PR #186, opened 2022-09-02
UI/TruncatingLabel

tonikocjan:
Implemented a UIView subclass called `TruncatingLabel` which renders two strings (called primary and secondary text).

The **rules** are the following:
1. Two lines of text can be displayed at the most.
2. The secondary text should not be truncated (it is truncated only when the line is too small for the whole string).
3. Primary text gets truncated if there is not enough space in two lines.
4. If there is enough space in one line for both texts, only allocate one line of space.

To get a better idea, here are the rules visualized:

```swift
// 1. Both strings fit into a single line
//  ---------------------------------------------
// |   This is a short text.   secondary label   |
//  ---------------------------------------------
//
// 2. Strings do not fit into a single line, however,
// there is enough space for both of them in two lines.
//  ---------------------------------------------
// |   This is a longer single-line text.        |
// |                           secondary label   |
//  ---------------------------------------------
//
//  ---------------------------------------------
// |   This is a longer two-line text, lorem     |
// |   ipsum dolor.            secondary label   |
//  ---------------------------------------------
//
// 3. Strings do not fit into a single line and there is also
// not enough space for the primary string in the second line.
// Primary string, therefore, has to be truncated with '...'.
//  ---------------------------------------------
// |   This is a long text that has to be split  |
// |   into two lines, ho...    secondary label  |
//  ---------------------------------------------
//
// 4. There is not enough space for the secondary text
// in the whole line. Truncate primary text in the first line.
//  ---------------------------------------------
// |   This is a long text that has to be sp...  |
// |   secondary label is too long to be disp... |
//  ---------------------------------------------
```

You are able to control fonts and text colors, and the direction the secondary string gravitates toward by modifying the `gravity` property. The default value is `left`:

```swift
  //  -------------------------------------------
  // |   This is a short text. secondary label   |
  //  -------------------------------------------
  //  -------------------------------------------
  // |   This is a longer single-line text.      |
  // |   secondary label                         |
  //  -------------------------------------------
  //  -------------------------------------------
  // |   This is a longer two-line text, lorem   |
  // |   ipsum dolor. secondary label            |
  //  -------------------------------------------
  //  -------------------------------------------
  // |   This is a longer two-line text, lorem   |
  // |   ipsum dolor lorem i... secondary label  |
  //  -------------------------------------------
```

The implementation does not use auto-layout (it relies on manual drawing) but can be used in an auto-layout context.

For testing, I (experimentally) used https://github.com/pointfreeco/swift-snapshot-testing package, which works by generating snapshots (images) of the view/view controller and comparing them the next time you run the test. It works well, but there is the drawback that you have to manually validate the images the first time you generate them, and when something does not work, you again have to check it manually. And I still have to make it work with GitHub actions.
You guys can tell me what you think about this and if you have another idea to test this, let me know.

### For future work (would be nice to have):

**Make it more configurable:**
1. spacings.
2. the number of lines (currently fixed to two).
3. ...


Sources/UI/TruncatingLabel/TruncatingLabel.swift
```diff
@@ -0,0 +1,325 @@
...
+// |   into two lines, ho...    secondary label  |
+//  ---------------------------------------------
+//
+// 4. There is not enough space for the secondary text
+// in the whole line. Truncate primary text in the first line.
+//  ---------------------------------------------
+// |   This is a long text that has to be sp...  |
+// |   secondary label is too long to be disp... |
+//  ---------------------------------------------
+//
```
borut-t, line 42: A nice set of example within the inline docs 👌

Sources/UI/TruncatingLabel/TruncatingLabel.swift
```diff
@@ -0,0 +1,325 @@
...
+    .foregroundColor: UIColor.gray
+  ]
+  private var secondaryTextAttributes: [NSAttributedString.Key: Any] = [
+    .font: UIFont.systemFont(ofSize: 12),
+    .foregroundColor: UIColor.blue
+  ]
+  
+  public var primaryColor: UIColor {
+    get {
+      primaryTextAttributes[.foregroundColor]! as! UIColor
```
borut-t, line 61: Although we are sure atm, the values are non-nil, this could change in the future. I would change this property to return an optional value instead. Just for the sake of safety. What do you think?

Sources/UI/TruncatingLabel/TruncatingLabel.swift
```diff
@@ -0,0 +1,325 @@
...
+    get {
+      secondaryTextAttributes[.font]! as! UIFont
+    }
+    set {
+      secondaryTextAttributes[.font] = newValue
+      internalIntrinsicContentSize = calculateIntrinsicContentSize()
+      setNeedsDisplay()
+    }
+  }
+  public var gravity: Gravity = .left {
```
borut-t, line 97: I would add a getter method so the receiver knows exactly what is gravity applied to 🤷‍♂️ 
```swift
/// Gravity is a rule that specifies the sticking point for the secondary text in relation to the primary text. The default value is `left`.
public var gravity: Gravity = .left {
```

Sources/UI/TruncatingLabel/TruncatingLabel.swift
```diff
@@ -0,0 +1,325 @@
...
+  //  -------------------------------------------
+  // |  This is a longer single-line text.       |
+  // |                          secondary label  |
+  //  -------------------------------------------
+  //  -------------------------------------------
+  // |  This is a longer two-line text, lorem    |
+  // |  ipsum dolor.            secondary label  |
+  //  -------------------------------------------
+  enum Gravity {
+    case left, right
```
borut-t, line 313: Add a comment for each case as well.
```swift
/// `secondaryText` sticks to the end of the `primaryText`
case left
/// `secondaryText` sticks to the right edge of this view
case right
```

borut-t: Maybe rename `left/right` to `leading/trailing` if we are going to support RTL languages some day 🤷‍♂️. Don't have strong feelings about that though.

Sources/UI/TruncatingLabel/TruncatingLabel.swift
```diff
@@ -0,0 +1,325 @@
...
+      /// 1. draw the first line
+      primary1.draw(
+        at: .zero,
+        withAttributes: primaryTextAttributes
+      )
+      /// 2. draw the second line
+      secondaryString.draw(
+        at: .init(
+          x: secondaryStartPosition,
+          y: primarySize.height + offset + (primarySize.height - secondarySize.height)*0.5),
```
borut-t, line 247: Let's add a space before and after the multiplication operator.
```swift
secondarySize.height) * 0.5)
```

borut-t (commented):
Awesome work with this rather specific view 🥳 

I have a few nits written below as comments and a few notes here:
- although I really like the tests for comparing the images, I would rather not use them here. It's pretty flaky and there prone to errors.
- not sure about the reason behind SnapKit integration, I would rather go without. Simply because we didn't need it before and having it just for one component seems a bit redundant Thoughts?
- one really big thing that I would do before we merge this component is to not be limited to only two lines. If we can achieve this in a timely good fashion, it would be awesome. The current state is obviously really specific in terms of usability.

tonikocjan:
> although I really like the tests for comparing the images, I would rather not use them here. It's pretty flaky and there prone to errors.

In what way do you think it's prone to errors? Do you have an alternative in mind? 

> not sure about the reason behind SnapKit integration, I would rather go without. Simply because we didn't need it before and having it just for one component seems a bit redundant Thoughts?

SnapKit is added for tests only.

> one really big thing that I would do before we merge this component is to not be limited to only two lines. 

This is something I marked on a todo list already. But I would do it on a separate branch.

borut-t:
> > although I really like the tests for comparing the images, I would rather not use them here. It's pretty flaky and there prone to errors.
> 
> In what way do you think it's prone to errors? Do you have an alternative in mind?

Not that value could be nil, it can't be. But it could get to a situation when some other dev will update the code and will not pay attention to this.

> 
> > not sure about the reason behind SnapKit integration, I would rather go without. Simply because we didn't need it before and having it just for one component seems a bit redundant Thoughts?
> 
> SnapKit is added for tests only.

Still, do we want it?

> 
> > one really big thing that I would do before we merge this component is to not be limited to only two lines.
> 
> This is something I marked on a todo list already. But I would do it on a separate branch.

Anyway is a good way. But keep in mind that I would not like to merge this (base) branch before the feature is somehow complete. Just by adding multiline support, it would be greatly beneficial and more generic than what it is atm.

Thoughts?

PR #187, opened 2022-09-02
Chore/Measurements

borut-t:
We don't need to calculate conversion manually anymore. There is a great [entity](https://developer.apple.com/documentation/foundation/measurement) from Foundation framework that can help us out here.


markom-povio (approved):
Nice and simple

tonikocjan (approved):
🙌

PR #188, opened 2022-09-05
Chore/Public access in ASB

borut-t:
@tonikocjan any particular reason why we were using `open` access?


tonikocjan (approved):
We can probably mark `AttributedStringBuilder` as final as well.

gentique (approved):

PR #189, opened 2022-09-05
Chore/async delay tests

gentique:
Since PR #184 fixed issues with concurrency with a short delay, I tried making it 0, and it still produced the same test results.
Thus further decreasing the test duration from 6.8~ to 0.25~.

### DIFF
--- a/Tests/Tests/PromiseKit/PromiseTests.swift
+++ b/Tests/Tests/PromiseKit/PromiseTests.swift
@@ -826,7 +826,7 @@ extension PromiseTests {
   func testConcurrentDispatch1() {
     func next(_ idx: Int) -> Promise<()>? {
       guard idx < 20 else { return nil }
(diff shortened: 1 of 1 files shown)


borut-t (approved):
Awesome 🙌

Tests/Tests/PromiseKit/PromiseTests.swift
```diff
@@ -826,7 +826,7 @@ extension PromiseTests {
   func testConcurrentDispatch1() {
     func next(_ idx: Int) -> Promise<()>? {
       guard idx < 20 else { return nil }
-      return async((), delay: .random(in: 0.01...0.05))
```
tonikocjan, line 829: @gentique I would like to keep it random because that adds a bit of non-determinism in the order in which promises resolve. Can we just lower the range a bit more?

tonikocjan (approved):

PR #190, opened 2022-09-05
Chore/Final implementations

borut-t:


gentique (approved):
Whilst I like the performance gains that we get from this, we should keep in mind that if there is a case where we depend on an implementation that isn't abstracted, we won't be able to write tests for that.

borut-t:
> Whilst I like the performance gains that we get from this, we should keep in mind that if there is a case where we depend on an implementation that isn't abstracted, we won't be able to write tests for that.

Yes, and that's a great thing as well.

markom-povio (approved):

PR #192, opened 2022-09-06
Chore/Deprecate DataSource protocol

borut-t:
Removing obosolete Data Source protocols since we are moving away to diffable data source.


yllfejziu:
Maybe we consider deprecating the DataSource protocol since some projects are already using it?

borut-t:
> Maybe we consider deprecating the DataSource protocol since some projects are already using it?

That's not a bad idea considering there are still projects with old approach. I'll do that instead. Thanks for suggestion 👍

README.md
```diff
@@ -24,22 +24,22 @@
 ## Packages
 ### Core libraries
 
-| Utilities | Protocols |
-| :--- | :--- |
-| [AppVersionValidator](Sources/Core/Utilities/AppVersionValidator/AppVersionValidator.swift) | [Data Source](Resources/Core/Protocols/DataSource/) |
```
gentique, line 29: We should also probably delete the README since we don't reference it anymore.

gentique (approved):

PR #193, opened 2022-09-07
Chore/Remove deprecated code

borut-t:


gentique (commented):
Given that we are deprecating some API-s, should we consider going to a 2.0.0 version?

borut-t:
> Given that we are deprecating some API-s, should we consider going to a 2.0.0 version?

I don't think deprecation is big enough reason to switch major version. We'll get to 2.0.0 sooner or later.

borut-t:
> > > Given that we are deprecating some API-s, should we consider going to a 2.0.0 version?
> > 
> > 
> > I don't think deprecation is big enough reason to switch major version. We'll get to 2.0.0 sooner or later.
> 
> I was in the middle of editing my comment, here's what I wanted to add to my reasoning:
> 
> e.g. incompatibility with previous versions. Per the semantic versioning and our own guideliness, this fits the case. `MAJOR version when you make incompatible API changes ` `When we do a feature or a set of features that are breaking changes, or could break the previous release...`

I'm completely aware of the reasoning behind the semantic versioning. And to that part you are true, this is indeed a major change.
Thanks @gentique for clarification and the feedback 👌 .

PR #198, opened 2022-09-09
Feature/XCTAssert+URLRequest

gentique:
Testing and asserting URLRequest-s isn't very reliable. Firstly there's no clear indicator as to why two URLRequests are different, it gives a generic message when it fails: 
Screen Shot 2022-09-09 at 3 22 32 PM

--

The second one is that it does not compare the `httpBody` and therefore can give you a passing test when reality it isn't.
Screen Shot 2022-09-09 at 3 13 41 PM

--

If we actually test for the `httpBody` we can see a failing test:
Screen Shot 2022-09-09 at 3 14 21 PM

--

With the help of this extension we can find out specifically what is failing and therefore fix it quickly.
Screen Shot 2022-09-09 at 3 15 46 PM

### DIFF
--- a/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
@@ -10,11 +10,11 @@ import XCTest
 
 public extension XCTestCase {
   func XCTAssertEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
-    XCTAssertEqual(lhs, rhs, file: file, line: line)
+    XCTAssertEqual(lhs, rhs, "Expected URLRequest to be equal", file: file, line: line)
     guard let lhs = lhs, let rhs = rhs else { return }
-    XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
-    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
-    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "httpBody", file: file, line: line)
-    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "allHTTPHeaderFields", file: file, line: line)
+    XCTAssertEqual(lhs.url, rhs.url, "Expected url to be equal", file: file, line: line)
+    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "Expected httpMethod to be equal", file: file, line: line)
+    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "Expected httpBody to be equal", file: file, line: line)
+    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "Expected HTTPHeaderFields to be equal", file: file, line: line)
   }
 }
--- a/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
@@ -9,6 +9,16 @@
 import XCTest
 
 public extension XCTestCase {
+  /**
+   Use this method when you need to thoroughly assert URLRequest.
+   
+   Asserted properties:
+   - `Self`
+   - `url`
+   - `httpMethod`
+   - `httpBody`
+   - `allHTTPHeaderFields`
+   */
   func XCTAssertEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
     XCTAssertEqual(lhs, rhs, "Expected URLRequest to be equal", file: file, line: line)
     guard let lhs = lhs, let rhs = rhs else { return }
@@ -18,6 +28,16 @@ public extension XCTestCase {
     XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "Expected HTTPHeaderFields to be equal", file: file, line: line)
   }
   
+  /**
+   Use this method when you need to thoroughly assert URLRequest.
+   
+   Asserted properties:
+   - `Self`
+   - `url`
+   - `httpMethod`
+   - `httpBody`
+   - `allHTTPHeaderFields`
+   */
   func XCTAssertNotEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
     XCTAssertNotEqual(lhs, rhs, "Expected URLRequest not to be equal", file: file, line: line)
     guard let lhs = lhs, let rhs = rhs else { return }
--- a/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
@@ -17,4 +17,13 @@ public extension XCTestCase {
     XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "Expected httpBody to be equal", file: file, line: line)
     XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "Expected HTTPHeaderFields to be equal", file: file, line: line)
   }
+  
+  func XCTAssertNotEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertNotEqual(lhs, rhs, "Expected URLRequest not to be equal", file: file, line: line)
+    guard let lhs = lhs, let rhs = rhs else { return }
+    XCTAssertNotEqual(lhs.url, rhs.url, "Expected url not to be equal", file: file, line: line)
+    XCTAssertNotEqual(lhs.httpMethod, rhs.httpMethod, "Expected httpMethod not to be equal", file: file, line: line)
+    XCTAssertNotEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "Expected httpBody not to be equal", file: file, line: line)
+    XCTAssertNotEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "Expected HTTPHeaderFields not to be equal", file: file, line: line)
+  }
 }
--- a/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
@@ -1,0 +1,20 @@
+//
+//  XCTestCase+PovioKit.swift
+//  PovioKit
+//
+//  Created by Gentian Barileva on 09/09/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import XCTest
+
+public extension XCTestCase {
+  func XCTAssertEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertEqual(lhs, rhs, file: file, line: line)
+    guard let lhs = lhs, let rhs = rhs else { return }
+    XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
+    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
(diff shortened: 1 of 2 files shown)


Sources/Core/Extensions/Other/XCTestCase+URLRequest.swift
```diff
@@ -0,0 +1,19 @@
...
+//  PovioKit
+//
+//  Created by Gentian Barileva on 09/09/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import XCTest
+
+public extension XCTestCase {
+  func XCTAssertEqualURLRequests(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
```
tonikocjan, line 12: Maybe we can override the existing ‘XCAssertEqual’ instead? I would also rather see non-optional arguments.

gentique: > Maybe we can override the existing ‘XCAssertEqual’ instead? I would also rather see non-optional arguments.

Hey Toni, Thanks for the feedback.

I tried overloading the method, but I'm getting some errors. Could you give me an example of how it would look like? It would greatly help me out.

Regarding the optional arguments, I have two points that I would like to mention:

The first one would be that sending optional values is already supported by XCTest methods, so it wouldn't fall outside their practices, and the second one is more related to how this method came to life.
The particular use case was for testing a URLSession client. Since it accepts URLRequest, I had to intercept the request and assert that it was the correct one. By that flow, if the implementation is wrong, the URLRequest can be optional. You can argue that you can assert unwrap the optional value, but I deemed it unnecessary since you can already XCTAssert optional and non-optional values.

tonikocjan: Let's keep optional arguments, I agree.

> Could you give me an example of how it would look like

This compiles for me:

```swift
func XCTAssertEqual(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
  XCTAssertTrue(lhs == rhs)
  guard let lhs = lhs, let rhs = rhs else { return }
  XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
  XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
  XCTAssertEqual(lhs.httpBody, rhs.httpBody, "httpBody", file: file, line: line)
  XCTAssertEqual(lhs.allHTTPHeaderFields, rhs.allHTTPHeaderFields, "allHTTPHeaderFields", file: file, line: line)
}
```

Sources/Core/Extensions/Other/XCTestCase+URLRequest.swift
```diff
@@ -0,0 +1,19 @@
...
+    XCTAssertEqual(lhs, rhs, file: file, line: line)
+    guard let lhs = lhs, let rhs = rhs else { return }
+    XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
+    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
+    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "httpBody", file: file, line: line)
+    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "allHTTPHeaderFields", file: file, line: line)  }
```
tonikocjan, line 18: You can omit the default values (`Data()` and `[:]`).

Sources/Core/Extensions/Other/XCTestCase+URLRequest.swift
gentique: If I try to add it as an extension to XCTestCase then it gives errors such as: 
`Use of 'XCTAssertEqual' refers to instance method rather than global function 'XCTAssertEqual(_:_:_:file:line:)' in module 'XCTest'`

Sources/Core/Extensions/Other/XCTestCase+URLRequest.swift
gentique: The reason why I defaulted to Empty Data and Dictionary is that it happens that URLRequest sets nil values as empty, and this can mess up with testing.
Screen Shot 2022-09-12 at 2 44 22 PM

Sources/Core/Extensions/Other/XCTestCase+URLRequest.swift
tonikocjan: Just make it a global function (other `XCTest...` methods are global as well).

tonikocjan, line 12: Should we also convert arguments to `@autoclosure`, to follow the XCTest design?

gentique: I tested this out as follows:
```swift
func XCTAssertEqual(_ lhs: @autoclosure() -> URLRequest?, _ rhs: @autoclosure() -> URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
  XCTAssertTrue(lhs() == rhs(), file: file, line: line)
  guard let lhs = lhs(), let rhs = rhs() else { return }
  XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
  XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
  XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "httpBody", file: file, line: line)
  XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "allHTTPHeaderFields", file: file, line: line)
}
```
But since the original XCTAssertEqual is of generic type I think it's being prioritized instead of our function.

borut-t, line 12: I would rename to singularity `XCTAssertEqualURLRequest`.

borut-t (approved):
Great 👌

Sources/Core/Extensions/Other/XCTestCase+URLRequest.swift
```diff
@@ -0,0 +1,20 @@
...
+//  PovioKit
+//
+//  Created by Gentian Barileva on 09/09/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import XCTest
+
+public extension XCTestCase {
+  func XCTAssertEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
```
borut-t, line 12: Let's put this under `XCTestCase+PovioKit` file instead.

borut-t (approved):

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -0,0 +1,20 @@
...
+
+public extension XCTestCase {
+  func XCTAssertEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertEqual(lhs, rhs, file: file, line: line)
+    guard let lhs = lhs, let rhs = rhs else { return }
+    XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
+    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
+    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "httpBody", file: file, line: line)
+    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "allHTTPHeaderFields", file: file, line: line)
+  }
```
borut-t, line 19: Can we make tests for this method as well?

gentique:
@borut-t Please check my latest commits, I added the same file as you did on #199, this way you can easily rebase and add your tests.

I've added test with various combinations to assert that this extension works exactly the same as `XCTAssertEqual`.
Coverage is at 100% for this extension. 
Screen Shot 2022-09-15 at 10 55 24 PM

Tests/Tests/Core/Extensions/Other/XCTestCaseTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    var secondUrlRequest = anyRequest()
+    secondUrlRequest.httpBody = data2
+    
+    // Test passes even though httpBody is different.
+    XCTAssertEqual(firstUrlRequest, secondUrlRequest)
+    
+    XCTExpectFailure("Expected test to fail since httpBody is different.") {
+      XCTAssertEqual(firstUrlRequest.httpBody, secondUrlRequest.httpBody)
+      XCTAssertEqualURLRequest(firstUrlRequest, secondUrlRequest)
+    }
```
borut-t, line 28: Nice use of block-based expectation 👌

Tests/Tests/Core/Extensions/Other/XCTestCaseTests.swift
```diff
@@ -0,0 +1,78 @@
...
+  }
+  
+  func test_XCTAssertEqualURLRequest_hasExpectedFailures() {
```
borut-t, line 31: Do you prefer snake case over a camel case?

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -0,0 +1,20 @@
...
+//  PovioKit
+//
+//  Created by Gentian Barileva on 09/09/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import XCTest
+
+public extension XCTestCase {
+  func XCTAssertEqualURLRequest(_ lhs: URLRequest?, _ rhs: URLRequest?, file: StaticString = #filePath, line: UInt = #line) {
```
borut-t, line 12: Do we want to have the `NotEqual` method as well? 🤔

borut-t (approved):
Great work @gentique. Added a few nits below. I'll rebase my branch once this gets merged.

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
gentique: I'm not sure honestly, maybe we can add it later on if such a need appears.

What do you think?

borut-t: It is useful in some cases. And to be consistent with XCTest framework, maybe we should add it 🤷‍♂️ . I don't have strong feelings about it though.

Tests/Tests/Core/Extensions/Other/XCTestCaseTests.swift
gentique: Hmm, this naming convention is something I got from the Essential Developer course. I like it since it makes it very clear what is being tested and what is being expected from the test without having to go into the test itself.

borut-t: I don't mind about the naming at all. Just the case type.
```swift
func test_XCTAssertEqualURLRequest_hasExpectedFailures()
vs 
func testXCTAssertEqualURLRequestHasExpectedFailures()
```
Again, don't have strong feelings. Maybe we should vote about it 😄

gentique: Sorry, I meant just what you said:

test_`System under test`_`Expected behaviour` the underline makes it easier to read in my opinion 🤷‍♂️ , maybe I just got used to it by now. We could vote on it on our next pod meeting 😄

borut-t: Maybe it's just me that I'm not used to this snake case. But I get your point. It is easier to read. And for tests, it's ok I think 👌 We can still do a quick vote on this at the next pod meeting 😄

borut-t:
@gentique just run tests and all your tests are failing 🤔

gentique:
> @gentique just run tests and all your tests are failing 🤔

Well yes, the way I have written these tests is to Assert that they correctly fail just like XCTAssert does, that's why they are wrapped in `XCTExpectFailure`. 

in cases where you have different http method, body, headers, you need to assert that the extension fails (this is the expected behaviour).

Screen Shot 2022-09-16 at 11 41 21 AM
You can see in the above tests that it does not show up as an actual failure.

I will rewrite this since upon making these tests "pass" (which is not expected) they don't report as failing.

borut-t:
> > @gentique just run tests and all your tests are failing 🤔
> 
> Well yes, the way I have written these tests is to Assert that they correctly fail just like XCTAssert does, that's why they are wrapped in `XCTExpectFailure`.
> 
> in cases where you have different http method, body, headers, you need to assert that the extension fails (this is the expected behaviour).
> 
> Screen Shot 2022-09-16 at 11 41 21 AM
> 
> You can see in the above tests that it does not show up as an actual failure. --
> #### I will rewrite this since upon making these tests "pass" (which is not expected) they don't report as failing.

My bad, they succeed indeed.

gentique:
- Added `XCTAssertNotEqualURLRequest`
  - Add tests
- Refactored tests
- Add documentation

markom-povio (approved):
Nice 👍

PR #199, opened 2022-09-15
Feature/XCTest extension methods

borut-t:
Similar to @gentique #198, added methods to compare images and fonts.

### DIFF
--- a/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
@@ -46,4 +46,34 @@ public extension XCTestCase {
     XCTAssertNotEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "Expected httpBody not to be equal", file: file, line: line)
     XCTAssertNotEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "Expected HTTPHeaderFields not to be equal", file: file, line: line)
   }
+  
+  func XCTAssertEqualImage(_ lhs: UIImage?, _ rhs: UIImage?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertNotNil(lhs, "Left image is nil", file: file, line: line)
+    XCTAssertNotNil(rhs, "Right image is nil", file: file, line: line)
+    guard let imageData1 = lhs?.pngData(),
+          let imageData2 = rhs?.pngData() else { return }
+    XCTAssertTrue(imageData1.elementsEqual(imageData2), "Images doesn't equal!", file: file, line: line)
+  }
+  
+  func XCTAssertNotEqualImage(_ lhs: UIImage?, _ rhs: UIImage?, file: StaticString = #filePath, line: UInt = #line) {
+    guard let imageData1 = lhs?.pngData(),
+          let imageData2 = rhs?.pngData() else { return }
+    XCTAssertFalse(imageData1.elementsEqual(imageData2), "Images do equal!", file: file, line: line)
+  }
+  
+  func XCTAssertEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertEqual(lhs?.pointSize, rhs?.pointSize, "PointSize doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs?.fontName, rhs?.fontName, "FontName doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs?.familyName, rhs?.familyName, "FamilyName doesn't equal!", file: file, line: line)
+  }
+  
+  func XCTAssertNotEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertNotNil(lhs, "Left font is nil", file: file, line: line)
+    XCTAssertNotNil(rhs, "Right font is nil", file: file, line: line)
+    let pointSizeEquals = lhs?.pointSize == rhs?.pointSize
+    let fontNameEquals = lhs?.fontName == rhs?.fontName
+    let familyNameEquals = lhs?.familyName == rhs?.familyName
+    guard pointSizeEquals, fontNameEquals, familyNameEquals else { return }
+    XCTFail("Fonts do equal!", file: file, line: line)
+  }
 }
--- a/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
+++ b/Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
@@ -61,13 +61,13 @@ public extension XCTestCase {
     XCTAssertFalse(imageData1.elementsEqual(imageData2), "Images do equal!", file: file, line: line)
   }
   
-  func XCTAssertEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
+  func XCTAssertEqualFont(_ lhs: UIFont?, _ rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
     XCTAssertEqual(lhs?.pointSize, rhs?.pointSize, "PointSize doesn't equal!", file: file, line: line)
     XCTAssertEqual(lhs?.fontName, rhs?.fontName, "FontName doesn't equal!", file: file, line: line)
     XCTAssertEqual(lhs?.familyName, rhs?.familyName, "FamilyName doesn't equal!", file: file, line: line)
   }
   
-  func XCTAssertNotEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
+  func XCTAssertNotEqualFont(_ lhs: UIFont?, _ rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
     XCTAssertNotNil(lhs, "Left font is nil", file: file, line: line)
     XCTAssertNotNil(rhs, "Right font is nil", file: file, line: line)
     let pointSizeEquals = lhs?.pointSize == rhs?.pointSize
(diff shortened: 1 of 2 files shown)


Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -12,9 +12,38 @@ public extension XCTestCase {
...
+    XCTAssertEqual(lhs.url, rhs.url, "Url doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "HttpMethod doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "HttpBody doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "AllHTTPHeaderFields doesn't equal!", file: file, line: line)
+  }
+  
+  func XCTAssertEqualImage(_ lhs: UIImage?, _ rhs: UIImage?, file: StaticString = #filePath, line: UInt = #line) {
+    guard let image1 = lhs, let image2 = rhs,
+          let imageData1 = image1.pngData(),
+          let imageData2 = image2.pngData() else { return }
```
gentique, line 24: We ought to change this guard statement, since this test will pass if I use the method this way 

`XCTAssertEqualImage(UIImage(), nil)`

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -12,9 +12,38 @@ public extension XCTestCase {
...
+    XCTAssertTrue(imageData1.elementsEqual(imageData2), "Images doesn't equal!", file: file, line: line)
+  }
+  
+  func XCTAssertNotEqualImage(_ lhs: UIImage?, _ rhs: UIImage?, file: StaticString = #filePath, line: UInt = #line) {
+    guard let image1 = lhs, let image2 = rhs,
+          let imageData1 = image1.pngData(),
+          let imageData2 = image2.pngData() else { return }
```
gentique, line 31: Same here.

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -12,9 +12,38 @@ public extension XCTestCase {
...
+  }
+  
+  func XCTAssertEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertEqual(lhs?.pointSize, rhs?.pointSize, "PointSize doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs?.fontName, rhs?.fontName, "FontName doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs?.familyName, rhs?.familyName, "FamilyName doesn't equal!", file: file, line: line)
+  }
+  
+  func XCTAssertNotEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
+    guard let lhs = lhs, let rhs = rhs else { return }
```
gentique, line 42: Same here.

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -12,9 +12,38 @@ public extension XCTestCase {
...
+    let pointSizeEquals = lhs.pointSize == rhs.pointSize
+    let fontNameEquals = lhs.fontName == rhs.fontName
+    let familyNameEquals = lhs.familyName == rhs.familyName
+    guard pointSizeEquals, fontNameEquals, familyNameEquals else { return }
+    XCTFail("Fonts do equal!", file: file, line: line)
```
gentique, line 47: I would do asserts for each property that we are testing, otherwise we lose the ability to distinguish what properties are equal in this particular test.

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -12,9 +12,38 @@ public extension XCTestCase {
...
     guard let lhs = lhs, let rhs = rhs else { return }
-    XCTAssertEqual(lhs.url, rhs.url, "url", file: file, line: line)
-    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "httpMethod", file: file, line: line)
-    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "httpBody", file: file, line: line)
-    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "allHTTPHeaderFields", file: file, line: line)
+    XCTAssertEqual(lhs.url, rhs.url, "Url doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs.httpMethod, rhs.httpMethod, "HttpMethod doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs.httpBody ?? Data(), rhs.httpBody ?? Data(), "HttpBody doesn't equal!", file: file, line: line)
+    XCTAssertEqual(lhs.allHTTPHeaderFields ?? [:], rhs.allHTTPHeaderFields ?? [:], "AllHTTPHeaderFields doesn't equal!", file: file, line: line)
+  }
```
gentique, line 19: Please add a comment on my PR regarding this change, since it's still open I'm more than happy to make it more descriptive.

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
```diff
@@ -12,9 +12,38 @@ public extension XCTestCase {
...
+  
+  func XCTAssertNotEqualImage(_ lhs: UIImage?, _ rhs: UIImage?, file: StaticString = #filePath, line: UInt = #line) {
+    guard let image1 = lhs, let image2 = rhs,
+          let imageData1 = image1.pngData(),
+          let imageData2 = image2.pngData() else { return }
+    XCTAssertFalse(imageData1.elementsEqual(imageData2), "Images do equal!", file: file, line: line)
+  }
+  
+  func XCTAssertEqualFont(lhs: UIFont?, rhs: UIFont?, file: StaticString = #filePath, line: UInt = #line) {
```
gentique, line 35: Let's omit the argument labels like in other methods.

Tests/Tests/Core/Extensions/Other/XCTestCaseTests.swift
```diff
@@ -0,0 +1,32 @@
...
+  func testEqualFont() {
+    XCTAssertEqualFont(lhs: UIFont.systemFont(ofSize: 12), rhs: UIFont.systemFont(ofSize: 12))
+  }
+  
+  func testNotEqualFont() {
+    XCTAssertNotEqualFont(lhs: UIFont.systemFont(ofSize: 12), rhs: UIFont.systemFont(ofSize: 13))
+    XCTAssertNotEqualFont(lhs: UIFont.systemFont(ofSize: 12), rhs: UIFont.boldSystemFont(ofSize: 12))
+    XCTAssertNotEqualFont(lhs: UIFont.systemFont(ofSize: 12), rhs: UIFont.italicSystemFont(ofSize: 12))
+  }
+}
```
gentique, line 32: I would add a new assertion for each test where we have optional values as arguments, that way we can further assert that we are covering all cases.

For example :
``` swift
XCTAssertNotEqualImage(UIImage(), nil)
XCTAssertNotEqualFont(lhs: UIFont.systemFont(ofSize: 12), rhs: nil)
```

gentique (changes requested):

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
borut-t: Correct 👌

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
borut-t: Not really. Comparing an image with nil should succeed not fail.

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
borut-t: Yes

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
borut-t: Not really. If you have the same font, only pointSize differs. Asserting to no equal pointSize will succeed, but the other two asserts will fail, therefore test will fail.

```swift
let lhs: UIFont? = UIFont.systemFont(ofSize: 12)
let rhs: UFont? = UIFont.systemFont(ofSize: 13)
XCTAssertNotEqual(lhs?.pointSize, rhs?.pointSize, "PointSize equals!", file: file, line: line) - succeeds
XCTAssertNotEqual(lhs?.fontName, rhs?.fontName, "FontName equals!", file: file, line: line) - fails
XCTAssertNotEqual(lhs?.familyName, rhs?.familyName, "FamilyName equals!", file: file, line: line) - fails
```

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
borut-t: We actually need them, otherwise error is not propagated forward.

Tests/Tests/Core/Extensions/Other/XCTestCaseTests.swift
borut-t: Yes

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
gentique: I am talking about `lhs: UIFont?, rhs: UIFont?` -> `_ lhs: UIFont?, _ rhs: UIFont?` not `file: StaticString = #filePath, line: UInt = #line` , or am I missing something?

borut-t: Missed that one, thanks 👌

Sources/Core/Extensions/Other/XCTestCase+PovioKit.swift
gentique: Ahh, so the idea is that if only one of those properties is different then we mark the fonts as different.

gentique (approved):

gentique:
👌

markom-povio (approved):

PR #200, opened 2022-09-19
Chore/More unit tests

borut-t:


Sources/Core/Utilities/DispatchTimer/DispatchTimer.swift
```diff
@@ -49,5 +51,11 @@ public extension DispatchTimer {
   func stop() {
     timer?.cancel()
     timer = nil
+    isExecuting = false
+  }
+  
+  var isActive: Bool {
+    guard let timer = timer else { return false }
+    return !timer.isCancelled && isExecuting
```
gentique, line 59: Could we simplify this by only checking if the timer is not nil & not cancelled? From what I noticed, we set isExecuting to true only when we instantiate the timer.

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
```diff
@@ -0,0 +1,133 @@
...
+  }
+  
+  func test_applyBorder() {
+    let color: UIColor = .blue
+    let width: CGFloat = 10
+    
+    let view = UIView()
+    view.layer.borderColor = UIColor.black.cgColor
+    XCTAssertEqual(view.layer.borderColor, UIColor.black.cgColor)
+    XCTAssertEqual(view.layer.borderWidth, 0)
```
gentique, line 52: This feels like we are testing the UIKit framework instead of our extension, maybe we should make the assumption that this always works, except if we have business logic that relies on this.

If you agree with this, I would suggest removing such assertions in other places too.

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
```diff
@@ -0,0 +1,133 @@
...
+  }
+  
+  func test_setHidden_true() {
+    let view = UIView()
+    
+    XCTAssertFalse(view.isHidden)
+    XCTAssertEqual(view.alpha, 1)
+    
+    let promise = expectation(description: "Wait for animation...")
+    view.setHidden(true, animationDuration: 0.1) {
```
gentique, line 66: Might sound overkill but could we shorten the delay duration?

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
```diff
@@ -0,0 +1,133 @@
...
+  }
+  
+  func test_rotate_clockwise() {
+    let view = UIView()
+    
+    view.rotate(speed: 1, clockwise: true)
+    let animation = view.layer.animation(forKey: UIView.AnimationKey.rotation) as? CABasicAnimation
+    XCTAssertNotNil(animation)
+    XCTAssertEqual(animation?.fromValue as? Int, 0)
+    XCTAssertEqual(animation?.toValue as? Float, (Float.pi * 2.0) * 1)
```
gentique, line 99: We should avoid magic numbers.

Tests/Tests/Core/Utilities/DispatchTimer/DispatchTimerTests.swift
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+    
+    XCTAssertTrue(timer.isActive)
+    waitForExpectations(timeout: 0.5)
+    XCTAssertFalse(timer.isActive)
+  }
+  
+  func test_dispatchTimer_static_noRepeating() {
+    let promise = expectation(description: "Wait for timer...")
```
gentique, line 49: Maybe we could add `promise.expectedFulfillmentCount = 1` like in previous test to make it more assertive what we are trying to achieve.

Sources/Core/Utilities/DispatchTimer/DispatchTimer.swift
borut-t: I've found out, relying only on `isCancelled`, tests were flaky. I'll check if this is still the case.

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
borut-t: That's like a precondition so we know for sure how was before and how it is now.

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
borut-t: There is no magic in this number, pure math :)

Tests/Tests/Core/Utilities/DispatchTimer/DispatchTimerTests.swift
borut-t: That feels redundant. We all know the default value is 1, so why overriding defaults 🤷‍♂️

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
gentique: > There is no magic in this number, pure math :)

Do you not agree that it would be nicer to declare them in a property, that way we know immediately what they represent?

`let expectedStartAngle = 0`
`let expectedEndAngle = (Float.pi * 2.0) * 1`

Tests/Tests/Core/Utilities/DispatchTimer/DispatchTimerTests.swift
gentique: > That feels redundant. We all know the default value is 1, so why overriding defaults 🤷‍♂️

The way you tested UIViews with the preconditions is also a bit redundant though you did it because you felt that they provided value, that's how I saw them here as well. Simply a bit easier to read for someone looking at them for the first time.

borut-t: I agree, it is. But that's how I felt it should be 🤷‍♂️

Tests/Tests/Core/Extensions/UIKit/UIViewTests.swift
borut-t: That would be a nice addition yes, but not necessary.

gentique (approved):
👌

yllfejziu (approved):
👍

PR #202, opened 2022-09-20
Chore/Delete obsolete method

borut-t:
@nejcvivod I think you are the author of this extension method. Do you want to retain (and fix) it or should we proceed and delete it?


nejcvivod:
@borut-t nuke it :)

nejcvivod (approved):

PR #204, opened 2022-09-22
Feature/New extensions

borut-t:


Sources/Core/Extensions/Foundation/String+Povio.swift
```diff
@@ -37,4 +37,14 @@ public extension String {
...
+  
+  /// Returns first n characters from string
+  func first(n characters: UInt) -> String {
+    .init(prefix(min(count, Int(characters))))
+  }
+  
+  /// Returns last n characters from string
+  func last(n characters: UInt) -> String {
+    .init(suffix(min(count, Int(characters))))
+  }
```
tonikocjan, line 49: These two methods feel superfluous as you can just use `prefix` / `suffix` directly. What value do you think they provide?

Sources/Core/Extensions/Foundation/Dictionary.swift
```diff
@@ -0,0 +1,25 @@
...
+import Foundation
+
+public extension Dictionary {
+  /// Make union from `Self` and given `dictionary`
+  /// Examples:
+  /// ```
+  /// print([1: "One", 2: "Two"].union([3: "Three", 4: "Four"]) // result: [1: "One", 2: "Two", 3: "Three", 4: "Four"]
+  /// print([1: "One", 2: "Two"].union([2: "Two"]) // result: [1: "One", 2: "Two"]
+  /// ```
+  func union(_ dictionary: Dictionary) -> Dictionary {
```
tonikocjan, line 18: I believe Dictionary already has something similar implemented:

- https://developer.apple.com/documentation/swift/dictionary/merge(_:uniquingkeyswith:)-7smbb
- https://developer.apple.com/documentation/swift/dictionary/merging(_:uniquingkeyswith:)-3vtfs

Sources/Core/Extensions/Foundation/String+Povio.swift
tonikocjan: > Prefix/suffix allows you to put negative values inside

That's a valid point, actually! It's one of the weird things about swift and integral types. For some reason, swift prefers `IntX` over `UIntX` even when the latter makes more sense (for instance, `collection.count`).

If we decide to use such safe versions, I would suggest a different naming, smth like `safePrefix(_ maxLength: UInt)` and `safeSuffix(_ maxLength: UInt)`, which tells the meaning clearly. Second, I would return a view into the string instead of a new copy. That increases the performance when a copy is not needed and leaves that option to the user:

```swift
 /// Returns first n characters from string
  func prefix(_ maxLength: UInt) -> Substring {
    prefix(Int(maxLength))
  }

  /// Returns last n characters from string
   func suffix(_ maxLength: UInt) -> Substring {
    suffix(Int(maxLength))
  }
```

Sources/Core/Extensions/Foundation/Dictionary.swift
borut-t: It's been there since iOS 8 🤦‍♂️ 
Thanks, Toni. Way better yes since it handles merges with different values.

markom-povio (approved):
Looking good

Sources/Core/Extensions/Foundation/Double+Povio.swift
```diff
@@ -9,15 +9,31 @@
...
+  
+  /// Converst `miles` to `meters`
+  var meters: Double {
+    Measurement(value: self, unit: UnitLength.miles)
+      .converted(to: .meters)
+      .value
+  }
+  
+  /// Converts `meters` to `miles`
+  var miles: Double {
```
gentique, line 34: I am thinking about this name and how it would look like in code, and I don't know how readable it would be at first sight:

```swift
(100.7).meters
```
My thoughts are, is this a Number which is now of type `meters`? Don't know, maybe I'm overthinking this 😆

gentique (approved):

Sources/Core/Extensions/Foundation/Double+Povio.swift
tonikocjan: It does not happen often that you would convert a literal like this. More often than not you would use it on a name, so the syntax is 

```swift
let miles: Double = ...
let meters = miles.meters
```

A bit more concerning to me is that `.meters` does not tell me _from_ which unit this converts (is it centimeters, kilometers, ...).

borut-t: > I am thinking about this name and how it would look like in code, and I don't know how readable it would be at first sight:
> 
> ```swift
> (100.7).meters
> ```
> 
> My thoughts are, is this a Number which is now of type `meters`? Don't know, maybe I'm overthinking this 😆

I think that is a mostly subjective point since brackets or no brackets, that's just syntactic sugar that helps you or the compiler to even some hidden problems.

```swift
(100.7).meters
vs
convertToMeters(100.7)
```
Looking at the example above, the second one with function doesn't seem any more readable to me. Again, subjective point. Lmk if you disagree or you have a better objective reason this would not work.

borut-t: > It does not happen often that you would convert a literal like this. More often than not you would use it on an identifier, so the syntax is
> 
> ```swift
> let miles: Double = ...
> let meters = miles.meters
> ```
> 
> A bit more concerning to me is that `.meters` does not tell me _from_ which unit this converts (is it centimeters, kilometers, ...).

I think we cannot discuss usability here since it all depends on the nature of the domain you are working on.

The second thought here is a bit vague, but I understand your point. I would say that this is irrelevant for generic conversions where you know what you are converting. When this is not the case, you would extrapolate value to a certain type anyway and do the conversion afterward. Lmk if you agree/disagree. I'm open to a discussion.

gentique: Here's an example of how I thought about this one:
```swift
public struct Meter {
 var value: Double
}

extension Meter: ExpressibleByIntegerLiteral {
  public init(integerLiteral value: IntegerLiteralType) {
      self.init(value: Double(value))
  }
}

public extension Meter {
  let asMiles: Double {
    Measurement(value: self.value, unit: UnitLength.meters)
      .converted(to: .miles)
      .value
  }
}

// Usage

let meters: Meter = 70
let miles = meters.asMiles

```

This way the intention is clear about what type is being converted to where.
Though this was just an experimental thought and not to be taken too seriously 😄

borut-t: Though I like expressibility in this example, it seems a bit too much coding for such a trivial thing. And I completely understand your standpoint.

I've been thinking about this and maybe we should replace all this "conversions" with one generic method.
```swift
func convert(from: Dimension, to: Dimension) -> Double {
  Measurement(value: self, unit: from)
    .converted(to: to)
    .value
}

// examples
let temp = (double-value).convert(from: UnitAngle.radians, to: UnitAngle.degrees)
let length = (double-value).convert(from: UnitLength.miles, to: UnitLength.meters)
```

By this, we know exactly which input and output unit is. Thoughts?

markom-povio: I like this one, it is more readable than the previous implementation.

borut-t: https://github.com/poviolabs/PovioKit/pull/204/commits/661d61b104c596ec00cc9a91fe5847a29c69bb52

gentique: This is much better. One minor thing that is very much developer error is the fact that you can mix different Units.

from: UnitLength, to: UnitMass - though this is apparent in Apple's code too, and I couldn't figure out a generic way to constraint this 🤷‍♂️ .

LGTM 👌

borut-t: Yes, that's the only downside of this library. You can convert from any to any unit. It will actually return some value, but who knows what it means when you convert kilograms to miles 😆 
Nevertheless, it doesn't crash when you do that, though someone would expect it would throw an error, it just silently does its "job".

PR #210, opened 2022-10-19
Feature: [UI] Camera

nagavcindriqim:
A UI component that simplifies handling with camera. It should be able to do the following:

- [X] a class Camera that can be used as base for all other types of camera capture
- [X] a class CameraService for handling with permissions
- [X] a class PhotoCamera derived from Camera which supports taking photos
- [X] a class QRCodeScanner derived from Camera which supports scanning QR codes

### DIFF
--- a/Resources/UI/Camera/README.md
+++ b/Resources/UI/Camera/README.md
@@ -1,0 +1,78 @@
+# Camera
+
+A UI component that simplifies handling with camera. `PhotoCamera` & `QRCodeScanner`.
+
+### PhotoCamera Example:
+```swift
+let camera = PhotoCamera()
+func prepareCamera() {
+  try? camera.prepare()
+    
+  // add previewLayer to any view
+  view.layer.addSublayer(camera.previewLayer)
+  camera.previewLayer.frame = self.view.bounds
+    
+  // set camera delegate
+  camera.delegate = self
+    
+  // get authorization status
+  let status = await camera.getAuthorizationStatus()
+    
+  switch status {
+  case .authorized:
+    self.camera.startSession()
+  case .denied:
+    // handle denied status
+    break
+  }
+}
+```
+#### Photo Camera functions
+`func takePhoto()`
+
+We can also switch from front facing to back camera:
+
+`func changeCamera(position: CameraPosition)`
+
+
+#### Photo Camera delegate
+`func photoCameraDidTakePhoto(_ image: UIImage)`
+
+`func photoCameraDidTriggerError(_ error: Camera.Error)`
+
+
+
+### QRCodeScanner Example:
+```swift
+let scanner = QRCodeScanner()
+func prepareQRCodeSCanner() {
+  scanner.prepare()
+    
+  // add previewLayer to any view
+  view.layer.addSublayer(scanner.previewLayer)
+  scanner.previewLayer.frame = self.view.bounds
+    
+  // set camera delegate
+  scanner.delegate = self
+    
+  // get authorization status
+  let status = await scanner.getAuthorizationStatus()
+    
+  switch status {
+  case .authorized:
+    self. scanner.startSession()
+  case .denied:
+    // handle denied status
+    break
+  }
+}
+```
+
+#### QRCodeScanner delegate
+`func codeScanned(code: String, boundingRect: CGRect)`
+
+`func scanFailure()`
+
+
+
+
--- a/Resources/UI/Camera/README.md
+++ b/Resources/UI/Camera/README.md
@@ -6,8 +6,12 @@ A UI component that simplifies handling with camera. `PhotoCamera` & `QRCodeScan
 ```swift
 let camera = PhotoCamera()
 func prepareCamera() {
-  camera.prepare()
-    
+  do {
+    try await camera.prepare()
+  } catch {
+    // error handling
+  }
+
   // add previewLayer to any view
   view.layer.addSublayer(camera.previewLayer)
   camera.previewLayer.frame = self.view.bounds
@@ -36,18 +40,21 @@ We can also switch from front facing to back camera:
 
 
 #### Photo Camera delegate
-`func photoCameraDidTakePhoto(_ image: UIImage)`
 
-`func photoCameraDidTriggerError(_ error: Camera.Error)`
+`func photoCamera(photoCamera: PhotoCamera, didTakePhoto image: UIImage)`
 
-
+`func photoCamera(photoCamera: PhotoCamera, didTriggerError error: Camera.Error)`
 
 ### QRCodeScanner Example:
 ```swift
 let scanner = QRCodeScanner()
 func prepareQRCodeSCanner() {
-  scanner.prepare()
-    
+  do {
+    try await camera.prepare()
+  } catch {
+    // error handling
+  }
+
   // add previewLayer to any view
   view.layer.addSublayer(scanner.previewLayer)
   scanner.previewLayer.frame = self.view.bounds
@@ -72,7 +79,3 @@ func prepareQRCodeSCanner() {
 `func codeScanned(code: String, boundingRect: CGRect)`
 
 `func scanFailure()`
-
-
-
-
--- a/Resources/UI/Camera/README.md
+++ b/Resources/UI/Camera/README.md
@@ -6,7 +6,7 @@ A UI component that simplifies handling with camera. `PhotoCamera` & `QRCodeScan
 ```swift
 let camera = PhotoCamera()
 func prepareCamera() {
-  try? camera.prepare()
+  camera.prepare()
     
   // add previewLayer to any view
   view.layer.addSublayer(camera.previewLayer)
@@ -15,13 +15,13 @@ func prepareCamera() {
   // set camera delegate
   camera.delegate = self
     
-  // get authorization status
-  let status = await camera.getAuthorizationStatus()
+  // request authorization status
+  let granted = await camera.requestAuthorizationStatus()
     
-  switch status {
-  case .authorized:
+  switch granted {
+  case true:
     self.camera.startSession()
-  case .denied:
+  case false:
     // handle denied status
     break
   }
@@ -55,13 +55,13 @@ func prepareQRCodeSCanner() {
   // set camera delegate
   scanner.delegate = self
     
-  // get authorization status
-  let status = await scanner.getAuthorizationStatus()
+  // request authorization status
+  let granted = await scanner.requestAuthorizationStatus()
     
-  switch status {
-  case .authorized:
-    self. scanner.startSession()
-  case .denied:
+  switch granted {
+  case true:
+    self.scanner.startSession()
+  case false:
     // handle denied status
     break
   }
--- a/Sources/UI/Camera/Camera.swift
+++ b/Sources/UI/Camera/Camera.swift
@@ -23,7 +23,7 @@ public class Camera: NSObject {
   let sessionQueue = DispatchQueue(label: "com.poviokit.camera")
   public lazy var previewLayer = AVCaptureVideoPreviewLayer(session: session)
   public let cameraService: CameraPermissionProviding
-  public var cameraPosition: CameraPosition = .back
+  var cameraPosition: CameraPosition = .back
   
   init(with cameraService: CameraPermissionProviding = CameraService()) {
     self.cameraService = cameraService
@@ -58,11 +58,11 @@ public extension Camera {
       guard self.session.isRunning else { return }
       self.session.stopRunning()
     }
-    setTorch(on: false) // just in case but flashlight is automatically turned off when session is stopped
+    try? setTorch(on: false) // just in case but flashlight is automatically turned off when session is stopped
   }
   
-  func toggleTorch() {
-    setTorch(on: !(device?.isTorchActive ?? true))
+  func toggleTorch() throws {
+    try setTorch(on: !(device?.isTorchActive ?? true))
   }
 }
 
@@ -73,19 +73,15 @@ private extension Camera {
     previewLayer.videoGravity = AVLayerVideoGravity.resizeAspectFill
   }
   
-  func setTorch(on: Bool) {
+  func setTorch(on: Bool) throws {
     guard let device = device, device.hasTorch, device.isTorchAvailable else { return }
-    do {
-      try device.lockForConfiguration()
-      switch on {
-      case true:
-        try device.setTorchModeOn(level: AVCaptureDevice.maxAvailableTorchLevel)
-      case false:
-        device.torchMode = .off
-      }
-      device.unlockForConfiguration()
-    } catch {
-      Logger.debug("Could not lock the camera device")
+    try device.lockForConfiguration()
+    switch on {
+    case true:
+      try device.setTorchModeOn(level: AVCaptureDevice.maxAvailableTorchLevel)
+    case false:
+      device.torchMode = .off
     }
+    device.unlockForConfiguration()
   }
 }
--- a/Sources/UI/Camera/Camera.swift
+++ b/Sources/UI/Camera/Camera.swift
@@ -22,10 +22,10 @@ public class Camera: NSObject {
   // Communicate with the session and other session objects on this queue.
   let sessionQueue = DispatchQueue(label: "com.poviokit.camera")
   public lazy var previewLayer = AVCaptureVideoPreviewLayer(session: session)
-  public let cameraService: CameraServiceProtocol
+  public let cameraService: CameraPermissionProviding
   public var cameraPosition: CameraPosition = .back
   
-  init(with cameraService: CameraServiceProtocol = CameraService()) {
+  init(with cameraService: CameraPermissionProviding = CameraService()) {
     self.cameraService = cameraService
     super.init()
     configureComponents()
@@ -41,15 +41,6 @@ public extension Camera {
   enum CameraPosition {
     case back
     case front
-    
-    var devicePosition: AVCaptureDevice.Position {
-      switch self {
-      case .back:
-        return .back
-      case .front:
-        return .front
-      }
-    }
   }
   
   enum CameraAuthorizationStatus {
@@ -61,15 +52,6 @@ public extension Camera {
   enum MediaType {
     case video
     case audio
-    
-    var type: AVMediaType {
-      switch self {
-      case .video:
-        return .video
-      case .audio:
-        return .audio
-      }
-    }
   }
   
   enum Error: Swift.Error {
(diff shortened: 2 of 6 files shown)


Sources/UI/Camera/CameraService.swift
```diff
@@ -0,0 +1,54 @@
...
+//  PovioKit
+//
+//  Created by Ndriqim Nagavci on 13/10/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import AVFoundation.AVMediaFormat
+import PovioKit
+
+protocol CameraServiceProtocol: AnyObject {
```
gentique, line 12: We should omit the word `Protocol`.

Sources/UI/Camera/CameraService.swift
```diff
@@ -0,0 +1,54 @@
...
+      }
+    @unknown default:
+      Logger.debug("Audio device in an unknown state ...")
+    }
+  }
+  
+  /// Returns current camera authorization status
+  func authorizationStatus(forType mediaType: AVMediaType) -> AVAuthorizationStatus {
+    AVCaptureDevice.authorizationStatus(for: mediaType)
+  }
```
gentique, line 47: This method from the protocol doesn't add any true value, it is more of a middleman. 

I think it would be more beneficial if we make its protocol and the implementation public.

Let me try to explain my thought process, basically in order for the client to request access to camera permissions it needs to initialize the whole Camera object. It would be much faster if we could access the CameraService object directly and request permission from there. I don't know if a "Camera" should care if it has permissions, from my pov it only needs to be able to capture and deliver images. Concern wise I feel like they should be separated.

What do you think?

Sources/UI/Camera/PhotoCamera.swift
```diff
@@ -0,0 +1,116 @@
...
+public extension PhotoCamera {
+  func prepare() throws {
+    sessionQueue.async {
+      do {
+        try self.configureComponents()
+      } catch {
+        
+      }
+    }
+  }
```
gentique, line 38: I don't think `prepare` will throw.

Sources/UI/Camera/CameraService.swift
nagavcindriqim: I agree. Making `CameraService` public will have its benefits in case the user doesn't want to initialize the whole Camera object to request or check permission status.

ArlindDushi (commented):
Great work! Really like what you've done here. 

Some suggestions not related to the code here:
-  Since you're working on this part maybe it wouldn't be a bad idea to add video recording capabilities to the camera. 
- Also maybe have the Scanner not be specific to QR code but having the ability to choose what kind of scanner we want (qr, barcode etc.) 

Thoughts @poviolabs/ios-developers ?

Sources/UI/Camera/Camera.swift
```diff
@@ -0,0 +1,138 @@
...
+    case audio
+  }
+  
+  enum Error: Swift.Error {
+    case unavailable
+    case missingSession
+    case missingInput
+    case missingOutput
+    case invalidImage
+  }
```
borut-t, line 63: Can we spawn enums into a standalone extension file `Camera+PovioKit`?

Sources/UI/Camera/Camera.swift
```diff
@@ -0,0 +1,138 @@
...
+extension Camera.MediaType {
+  var asAVMediaType: AVMediaType {
+    switch self {
+    case .video:
+      return .video
+    case .audio:
+      return .audio
+    }
+  }
+}
```
borut-t, line 138: I would move this two extensions into `Camera+PovioKit` file as well. Thoughts?

Sources/UI/Camera/CameraService.swift
```diff
@@ -0,0 +1,59 @@
...
+      return .authorized
+    case .denied, .restricted:
+      return .denied
+    case .notDetermined:
+      return .notDetermined
+    @unknown default:
+      Logger.debug("Audio device in an unknown state ...")
+      return .denied
+    }
+
```
borut-t, line 51: Redundant new line.

Sources/UI/Camera/CameraService.swift
```diff
@@ -0,0 +1,59 @@
...
+  public func authorizationStatus(forType mediaType: Camera.MediaType) -> Camera.CameraAuthorizationStatus {
+    switch AVCaptureDevice.authorizationStatus(for: mediaType.asAVMediaType) {
+    case .authorized:
+      return .authorized
+    case .denied, .restricted:
+      return .denied
+    case .notDetermined:
+      return .notDetermined
+    @unknown default:
+      Logger.debug("Audio device in an unknown state ...")
```
borut-t, line 48: Should we change the level to error and modify the message?
```swift
Logger.error("Camera device in an unknown state!")
```

Sources/UI/Camera/PhotoCamera.swift
```diff
@@ -0,0 +1,116 @@
...
+public extension PhotoCamera {
+  func prepare() {
+    sessionQueue.async {
+      do {
+        try self.configureComponents()
+      } catch {
+        
+      }
+    }
+  }
```
borut-t, line 38: What happens on error?

Sources/UI/Camera/PhotoCamera.swift
```diff
@@ -0,0 +1,116 @@
...
+  
+  func changeCamera(position: CameraPosition) {
+    if cameraPosition == position {
+      return
+    }
+    cameraPosition = position
+    prepare()
```
borut-t, line 45: I think guard makes more sense here...
```swift
guard cameraPosition != position else { return }
...
```

Resources/UI/Camera/README.md
```diff
@@ -0,0 +1,78 @@
+# Camera
+
+A UI component that simplifies handling with camera. `PhotoCamera` & `QRCodeScanner`.
+
+### PhotoCamera Example:
+```swift
+let camera = PhotoCamera()
+func prepareCamera() {
+  camera.prepare()
```
borut-t, line 9: What's the purpose of `prepare` method we need to trigger manually? Should it be implicitly executed instead?

Resources/UI/Camera/README.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+#### QRCodeScanner delegate
+`func codeScanned(code: String, boundingRect: CGRect)`
+
+`func scanFailure()`
+
+
+
+
```
borut-t, line 78: Only one new line is permitted at the end of the file.

borut-t (changes requested):
Great work. I wrote a few comments and nits below, mostly cosmetic.

Sources/UI/Camera/PhotoCamera.swift
nagavcindriqim: I was thinking of adding a completion block
`func prepare(completion: ((Error?) -> Void)? = nil)`

This way we make sure the user knows if the configuration failed for any reason.
The other approach would be to log the error, but I would prefer adding completion. Thoughts?

Sources/UI/Camera/Camera.swift
nagavcindriqim: I agree, will move these to `Camera+PovioKit`

Sources/UI/Camera/CameraService.swift
```diff
@@ -0,0 +1,59 @@
...
+  public func requestCameraAuthorization() async -> Bool {
+    let mediaType = Camera.MediaType.video
+    let authStatus = authorizationStatus(forType: mediaType)
+
+    if authStatus == .notDetermined {
+      let granted = await AVCaptureDevice.requestAccess(for: mediaType.asAVMediaType)
+      return granted
+    }
+    
+    return authStatus == .authorized ? true : false
```
markom-povio, line 35: Can this be just `return authStatus == .authorized`?

markom-povio (approved):
Nice 👍

tonikocjan (changes requested):

borut-t:
> Great work! Really like what you've done here.
> 
> Some suggestions not related to the code here:
> 
> * Since you're working on this part maybe it wouldn't be a bad idea to add video recording capabilities to the camera.
> * Also maybe have the Scanner not be specific to QR code but having the ability to choose what kind of scanner we want (qr, barcode etc.)
> 
> Thoughts @poviolabs/ios-developers ?

Good point @ArlindDushi. I think the first point is valid. We should have this supported. But not as part of this ticket. Maybe in a follow-up ticket in the future.
Regarding the second point, I think we should not put too much effort in since QR scanning is quite common nowadays compared to other types like barcodes. We don't want to cover all the variants, but just those common ones.

Sources/UI/Camera/Camera.swift
```diff
@@ -0,0 +1,138 @@
...
+
+public class Camera: NSObject {
+  var device: AVCaptureDevice? {
+    switch cameraPosition {
+    case .back:
+      return cameraService.isCameraAvailable(position: .back) ? AVCaptureDevice.default(for: .video) : nil
+    case .front:
+      return cameraService.isCameraAvailable(position: .front) ? AVCaptureDevice.default(.builtInWideAngleCamera, for: .video, position: .front) : nil
+    }
+  }
```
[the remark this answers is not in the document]
borut-t: Good point! I would go for the latter since creating new instances is expensive probably.

Sources/UI/Camera/Camera.swift
```diff
@@ -0,0 +1,138 @@
...
+  func toggleTorch() {
+    setTorch(on: !(device?.isTorchActive ?? true))
+  }
+}
+
+// MARK: - Private Methods
+private extension Camera {
+  func configureComponents() {
+    guard cameraService.isCameraAvailable(position: cameraPosition) else { return }
+    previewLayer.videoGravity = AVLayerVideoGravity.resizeAspectFill
```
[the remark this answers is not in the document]
borut-t: Correct, this should only be a default value.

Sources/UI/Camera/CameraService.swift
```diff
@@ -0,0 +1,59 @@
...
+//
+
+import AVFoundation.AVMediaFormat
+import PovioKit
+
+public protocol CameraPermissionProviding  {
+  func requestCameraAuthorization() async -> Bool
+  func authorizationStatus(forType mediaType: Camera.MediaType) -> Camera.CameraAuthorizationStatus
+  func isCameraAvailable(position: Camera.CameraPosition) -> Bool
+}
```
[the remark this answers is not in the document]
borut-t: We don't 🤔

Sources/UI/Camera/PhotoCamera.swift
nagavcindriqim: Since `session.startRunning()` is a blocking call and `configureComponents` must be called before that, I think it's safer to have this also in the same thread.  We can make the method async.

Resources/UI/Camera/README.md
nagavcindriqim: Will change `prepare` to an async method so the user has feedback if something failed during configuration.

Sources/UI/Camera/QRCodeScanner.swift
```diff
@@ -0,0 +1,73 @@
...
+      
+      // add output
+      if session.canAddOutput(metadataOutput) {
+        session.addOutput(metadataOutput)
+        metadataOutput.setMetadataObjectsDelegate(self, queue: .main)
+        if metadataOutput.availableMetadataObjectTypes.contains(.qr) {
+          metadataOutput.metadataObjectTypes = [.qr]
+        }
+      }
+    }
```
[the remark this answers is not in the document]
nagavcindriqim: Good point. Will need to throw instead

Sources/UI/Camera/Camera.swift
```diff
@@ -13,22 +13,27 @@ public class Camera: NSObject {
...
   // Communicate with the session and other session objects on this queue.
   let sessionQueue = DispatchQueue(label: "com.poviokit.camera")
-  public lazy var previewLayer = AVCaptureVideoPreviewLayer(session: session)
-  public let cameraService: CameraPermissionProviding
-  public var cameraPosition: CameraPosition = .back
+  public lazy var previewLayer: AVCaptureVideoPreviewLayer = {
+    let pl = AVCaptureVideoPreviewLayer(session: session)
+    pl.videoGravity = AVLayerVideoGravity.resizeAspectFill
+    
+    return pl
```
borut-t, line 28: Try to avoid using abbreviations for variable names. Naming it `previewLayer` is much clearer.

Sources/UI/Camera/Camera.swift
```diff
@@ -58,34 +63,31 @@ public extension Camera {
...
 
 // MARK: - Private Methods
 private extension Camera {
-  func configureComponents() {
-    guard cameraService.isCameraAvailable(position: cameraPosition) else { return }
-    previewLayer.videoGravity = AVLayerVideoGravity.resizeAspectFill
+  /// Check if camera is available on device
+  func isCameraAvailable(position: Camera.CameraPosition) -> Bool {
+    let session = AVCaptureDevice.DiscoverySession(deviceTypes: [deviceType], mediaType: .video, position: position.asAVCaptureDevicePosition)
+    return !session.devices.isEmpty
```
borut-t, line 79:
```swift
!AVCaptureDevice
  .DiscoverySession(deviceTypes: [deviceType], mediaType: .video, position: position.asAVCaptureDevicePosition)
  .devices
  .isEmpty
```
🤷‍♂️

Sources/UI/Camera/PhotoCamera.swift
```diff
@@ -68,16 +60,32 @@ public extension PhotoCamera {
       self.photoOutput.capturePhoto(with: photoSettings, delegate: self)
     }
   }
+  
+  func takePhoto(with photoSettings: AVCapturePhotoSettings, videoOrientation: AVCaptureVideoOrientation? = nil) {
+    let videoOrientation = (videoOrientation != nil) ? videoOrientation : previewLayer.connection?.videoOrientation
+    sessionQueue.async {
+      guard self.session.isRunning else {
```
borut-t, line 67: Do we have a retain cycle here?

borut-t (approved):

Sources/UI/Camera/PhotoCamera.swift
nagavcindriqim: I ran this test on a real device and there doesn't seem to be a retain cycle
```
func test_sut_isDeallocated() async {
  var sut: PhotoCamera? = PhotoCamera()
  
  try? await sut!.prepare()
  sut!.startSession()
  sut!.takePhoto()
  
  sut = nil
  
  XCTAssertNil(sut)
}
  ```
  
  The test passed and also `Camera` was deinitialized

Comment (tonikocjan) [reply Sources/UI/Camera/PhotoCamera.swift:67]: There should be no problem with retaining because the queue will, at some point, execute the work item and remove a reference. But I don't understand the test here; you are setting a value of `Optional<PhotoCamera>` to `.none` and then checking if it is `.none`. Of course, this will always be the case. Or am I missing something?

Comment (nagavcindriqim) [reply Sources/UI/Camera/PhotoCamera.swift:67]: Looks like there is a retain cycle after running this:
```
func expectNoMemoryLeaks(for object: AnyObject, file: StaticString = #filePath, line: UInt = #line) {
  addTeardownBlock { [weak object] in
   XCTAssertNil(object, file: file, line: line)
  }
}
  ```

Sources/UI/Camera/PhotoCamera.swift
```diff
@@ -0,0 +1,124 @@
...
+//  Created by Ndriqim Nagavci on 13/10/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import AVFoundation
+import UIKit
+
+public protocol PhotoCameraDelegate: AnyObject {
+  func photoCamera(photoCamera: PhotoCamera, didTakePhoto image: UIImage)
+  func photoCamera(photoCamera: PhotoCamera, didTriggerError error: Camera.Error)
```
tonikocjan, line 14: The call side would be nicer if the argument labels for `photoCamera` is `_`.

tonikocjan (approved):
Nice work. Two minor things to look at.

Sources/UI/Camera/Camera.swift
tonikocjan:
```
!AVCaptureDevice
  .DiscoverySession(
    deviceTypes: [deviceType], 
    mediaType: .video, 
    position: position.asAVCaptureDevicePosition)
  .devices
  .isEmpty
```

🤷‍♀️

Sources/UI/Camera/Camera.swift
```diff
@@ -0,0 +1,98 @@
...
+  let sessionQueue = DispatchQueue(label: "com.poviokit.camera")
+  public lazy var previewLayer: AVCaptureVideoPreviewLayer = {
+    let previewLayer = AVCaptureVideoPreviewLayer(session: session)
+    previewLayer.videoGravity = AVLayerVideoGravity.resizeAspectFill
+    
+    return previewLayer
+  }()
+  public let cameraService: CameraService
+  public var deviceType: AVCaptureDevice.DeviceType = .builtInWideAngleCamera
+  var cameraPosition: CameraPosition = .back
```
borut-t, line 32: We have a setter `setCameraPosition`. Can make this readonly?

Sources/UI/Camera/PhotoCamera.swift
```diff
@@ -0,0 +1,128 @@
...
+      DispatchQueue.main.async { self.delegate?.photoCamera(self, didTriggerError: .invalidImage) }
+      return
+    }
+    DispatchQueue.main.async { self.delegate?.photoCamera(self, didTakePhoto: image) }
+  }
+}
+
+// MARK: - Private Methods
+private extension PhotoCamera {
+  func configureComponents() throws {
```
borut-t, line 98: Let's rename it to `configure` 🤷‍♂️

Sources/UI/Camera/QRCodeScanner.swift
```diff
@@ -0,0 +1,75 @@
...
+    
+    guard let codeFrame = previewLayer.transformedMetadataObject(for: scannedObject)?.bounds else { return }
+    delegate?.codeScanned(code: qrCode, boundingRect: codeFrame)
+  }
+}
+
+// MARK: - Private Methods
+private extension QRCodeScanner {
+  func configureComponents() throws {
```
borut-t, line 50: Should we rename to `configure`?

borut-t (approved):
Great. Let's wrap this up. I think the implementation is solid.

Sources/UI/Camera/Camera.swift
nagavcindriqim: `setCameraPosition` is part of `PhotoCamera`, and every time the camera position is changed we need to run `configureComponents` on `PhotoCamera` so I would not make this readonly

borut-t: My bad 👌

PR #213, opened 2022-10-26
ConcurrentlyDispatch/Return results

tonikocjan:


borut-t (approved):
I've tested this on a real project and it does what it advertises. Great work 🙌

markom-povio (approved):
Nice!

PR #217, opened 2022-11-11
Feature: [UI] ARView

nagavcindriqim:
A UI component that can be used to display and interact with virtual objects. 

It displays 3D objects using `VirtualObject` class which can be initialized from a SCNode.
Gestures included are Tap, Pan, Pinch, and Rotation.


borut-t (commented):
Nice implementation. Since I don't have enough domain experience I would urge @tonikocjan and @markom-povio to review it as well since they've been working with ARKit recently.

Regarding inclusion into the PovioKit, I have some doubts. I prefer this to be placed into the resources repo instead due to the sheer specifics of it. But prove me wrong. Open for discussion.

nagavcindriqim:
Thanks for the reviews @borut-t & @markom-povio. I agree this can serve as a starting point and an example.

borut-t:
@nagavcindriqim thanks for this nice AR-related code. Since this is quite a specific code I would move it to our resource center under Examples instead. Can you do that and close this PR? Thanks!

PR #219, opened 2022-11-14
Feature/Collection extension to group elements by date

yllfejziu:
It's not rare we have to separate sections on a tableview by date. The idea behind this implementation is that we would be able to easily group an array by date by specifying the date field when conforming to `Dated` protocol.

To get a visual grasp when this extension would be useful, please take a look at the screenshot below:

IMG_2075


Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
```diff
@@ -21,3 +21,18 @@ public extension Collection {
     lazy.filter(clause).count
   }
 }
+
+public extension Collection where Element: Dated {
+  /// Groups dated collection elements returning a dictionary
+  func grouped(by dateComponents: Set<Calendar.Component> = [.year, .month, .day]) -> [Date: [Element]] {
+    let initial: [Date: [Element]] = [:]
+    let groupedByDateComponents = reduce(into: initial) { acc, cur in
+      let components = Calendar.current.dateComponents(dateComponents, from: cur.date)
```
EgzonArifi, line 30: I believe we need to inject the Calendar as a parameter, default value can be .current. What do you think?

yllfejziu: Yes, I agree 👍

tonikocjan (changes requested):
Cool idea! Left a comment regarding performance.

However, I would like to see a more _general_ implementation - there is no reason to limit ourselves to grouping by date only. One idea is to provide a closure as an argument, which extracts the key from the element, and use that key for grouping:

```swift
public extension Collection {
  func grouped<T: Hashable>(
    extract key: (Element) -> T
  ) -> [T: [Element]] {
    reduce(into: .init()) { acc, el in // notice type-inference: I don't have to specify the type of `into`
      let k = key(el)
      // ... perform the logic

```

Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
```diff
@@ -21,3 +21,18 @@ public extension Collection {
...
+      let date = Calendar.current.date(from: components) ?? Date()
+      let existing = acc[date] ?? []
+      acc[date] = existing + [cur]
```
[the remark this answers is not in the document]
tonikocjan: Some benchmarks might be interesting to look at. 😉

Tests/Tests/Core/Extensions/Foundation/CollectionDatedTests.swift
```diff
@@ -0,0 +1,64 @@
...
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import XCTest
+import PovioKit
+
+final class CollectionDatedTests: XCTestCase {
+  func test_grouped_returnsEmptyDictionaryOnEmptyArray() {
+    let sut: [Mock] = []
+    XCTAssertTrue(sut.grouped().isEmpty)
```
gentique, line 15: Nit, it is preferable to use `XCTAssertEqual` since in the case of failure the reason why is much easier to infer. We could also leave it as is but with a failure message.

Tests/Tests/Core/Extensions/Foundation/CollectionDatedTests.swift
```diff
@@ -0,0 +1,64 @@
...
+    
+    for value in grouped.values {
+      XCTAssertEqual(value.count, 1)
+    }
+  }
+}
+
+// MARK: - Helpers
+extension CollectionDatedTests {
+  class Mock: Dated {
```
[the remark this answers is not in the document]
yllfejziu: Since we don't verify the order of the functions called , I agree that's the case.

borut-t (commented):
I like the so,lution and I like Toni's proposal about the generic solution.
I also have one nit regarding the output of this function. It returns dictionary, meaning you'll need to sort it out every time you want to present it. Should we go with a two-dimensional array instead for a result?

Here is a [gist](https://gist.github.com/ElegyD/abc5dc819c9e7286452a6eb7e92fa9bb) that I remember using on one of our projects in the past.

tonikocjan:
I think the implementation proposed in this PR is much better for two reasons: 1. it is much simpler and easier to reason about; 2. it is much faster. 
To show how much faster, I prepared a detailed benchmarking. The tables below show the results.

I benchmarked three methods:
 1. `grouped`: this is the original implementation from this PR
 2. `grouped++`: this is the original implementation from this PR together with the proposed optimization
 3. `groupedByDay`: the method proposed by @borut-t 

Benchmarks for `grouped` and `grouped++` also sort the resulting dictionaries, so that they are comparable to the `groupByDay` algorithm.

Each of the methods is benchmarked using three different workloads. For each workload, a different-sized input is created to see how the algorithms scale.

## Benchmarks

**Benchmark A**: randomly sampled dates (some overlay is expected):

| n | `grouped`   | `grouped++` | `groupedBy` |
|--|--|--|--|
| 1e3 |   0.0027   |   0.0027  |  0.0053  |
| 1e4 |   0.0546   |  0.0254   |  0.2291  |
| 1e5 |   2.8673   |  0.2554   |  2121.1  |
| 1e6 |   20.699   |   2.5961  |  -*  |
| 1e7 |   383.52   |   27.555  |   -* |

\* Would take too much time to get the results.

**Benchmark B**: always the same date (_this is expected to be worst-case input for`grouped`_):

| n | `grouped`   | `grouped++` | `groupedBy` |
|--|--|--|--|
| 1e3 |   0.0032  |   0.0028  |  0.0062  |
| 1e4 |   0.0665   |  0.0298   |  0.3987  |
| 1e5 |   2.8673   |  0.2867   |   40.4176  |
| 1e6 |   356.40   |   2.9225  |  -  |
| 1e7 |   67834* (18h)  |   30.603  |  -  |

\* Might not be very reliable because the test ran only once so large fluctuations could be happening We can still agree, though, that it is _a bit_ slower than the ++ version.

**Benchmark C**: always a different date (_this is expected to be the best-case for `groupedByDay`_):

| n | `grouped`   | `grouped++` | `groupedBy` |
|--|--|--|--|
| 1e3 |   0.0037   |   0.0035  |  0.0032  |
| 1e4 |   0.0656   |  0.0365  |  0.0431  |
| 1e5 |   3.1612   |  0.3729  |  1.5960  |
| 1e6 |   37.6298   |   2.5961  |  31.912  |
| 1e7 |   244.8349   |   38.627  |  -  |

## Analysis

For `grouped`, we can observe, that in most cases, it is around ~10 times slower than the ++ version. However, as shown by _benchmark B_, its worst-case performance is much lower. This happens because the line `existing + [cur]` has to allocate a new array (which in itself is a big overhead), copy all of the elements into it, then allocate another array, and copy its (single) element into the resulting array. This converts a linear algorithm into a quadratic one (in the worst case).

`groupByDay` performs, as expected, worst still. My observation is that it is O(n^3) algorithm, so not usable for anything larger than around `n=50k`. Notice that it outperforms `grouped` in the `benchmark C`. However, this is mostly because dates are generated increasingly. When changed to decreasing fashion, which is the worst-case for this algorithm, it again performs much slower. This does not affect the other two algorithms because they rely on hashing instead of linear search.

**Benchmark C'**: dates generated in a decreasing fashion:

| n | `groupedBy` |
|--|--|
| 1e3 | 0.007  |
| 1e4 | 0.4061  |
| 1e5 | 4307.9  |
| 1e6 |  -  |
| 1e7 |  -  |


---

The [code](https://gist.github.com/tonikocjan/ce1b5eadfb8a0c4a39f9e9a14fd7a430). You're encouraged to repeat the experiments and report your findings.

borut-t:
Thanks, @tonikocjan, for an effort in doing extensive testing 🙌 
Although performance is not critical to our use cases dealing only with a couple of dates, I like your notion here.

@yllfejziu, will you be willing to update the current solution with @tonikocjan's proposal?

tonikocjan:
> Although performance is not critical to our use cases dealing only with a couple of dates

I agree, but as this is a library you cannot forsee how the method will be used. Therefore, the implementation should be as solid as possible.

borut-t:
> I agree, but as this is a library you cannot forsee how the method will be used. Therefore, the implementation should be as solid as possible.

Looks like I've expressed my notion incorrectly. It's not critical, but I always welcome any performance improvements and gains 💪

tonikocjan:
@yllfejziu I've been looking into the standard library and noticed the following method: https://developer.apple.com/documentation/swift/dictionary/init(grouping:by:) 
At the first glance, it looks like it might be doing something similar to what you are proposing here.

yllfejziu:
Great effort @tonikocjan and nice insights 👏 
I didn't have the time to update the PR with your suggestions but I'll be doing that soon enough.

yllfejziu:
> @yllfejziu I've been looking into the standard library and noticed the following method: https://developer.apple.com/documentation/swift/dictionary/init(grouping:by:) At the first glance, it looks like it might be doing something similar to what you are proposing here.

@tonikocjan  - Yes I was aware of this when I started this PR, therefore that's why I made specific implementation regarding date and not a general one (which I actually tried but when I saw this `grouped` apple function was already existent, I thought it'd be just redundant). 
What do you think? Should I just cover those performance comments you left and keep the implementation with dates, or you still think we should go for a more general approach?

tonikocjan:
I think we do not need the general method that I proposed. Instead, I would re-implement the method using Apple's grouped implementation.

tonikocjan (approved):
LGTM. 👍

borut-t (approved):
Great 👌

gentique (approved):

EgzonArifi (approved):
👌

PR #220, opened 2022-11-23
Auth/Social providers

borut-t:
Introducing three new products `PovioKitAuthApple`, `PovioKitAuthGoogle` and `PovioKitAuthFacebook` which are used to support the most popular and frequently used social login providers. `SignInWithApple` was deprecated.

The structure of PovioKit is now like this:
```
Core
Networking
PromiseKit
UI
AuthCore
-- Apple
-- Goole
-- Facebook
```


Package.swift
```diff
@@ -10,12 +10,13 @@ let package = Package(
     .library(name: "PovioKit", targets: ["PovioKit"]),
     .library(name: "PovioKitNetworking", targets: ["PovioKitNetworking"]),
     .library(name: "PovioKitPromise", targets: ["PovioKitPromise"]),
-    .library(name: "PovioKitUI", targets: ["PovioKitUI"])
+    .library(name: "PovioKitUI", targets: ["PovioKitUI"]),
+    .library(name: "PovioKitAuthApple", targets: ["PovioKitAuthApple"]),
+    .library(name: "PovioKitAuthGoogle", targets: ["PovioKitAuthGoogle"]),
```
tonikocjan, line 15: Should we create a new submodule named `PovioKitAuth` and move these two inside?

borut-t: The reason for this decision was based on the fact, that we might have another provider in the future (Fb, Snapchat, Twitter, LinkedIn, etc). But good idea yes.

tonikocjan: Yep. I think it would be nicer if we have them all in a dedicated `Auth` submodule.

borut-t: Actually, we could go with that right now. And when a new provider came, we create a new product for that. So essentially, we'll have Apple in "core", and every other as separate product.

borut-t: > Yep. I think it would be nicer if we have them all in a dedicated `Auth` submodule.

The thing about having them all within the same product is dependencies. Imagine you only support Apple, but you'll have Facebook, Google, etc dependencies bundles within your app. Not good.

borut-t: Good point Toni. I think that scenario, where the app would support all providers is not impossible, but rather less frequent. I think your solution is doable, but not really sure about the usability. I would rather have smth like this:
```
Auth (core module with core features and Apple provider, since there is no dependency needed)
  - Google (google provider with dependency on core module)
  - Fb (FB provider with  dependency on core module)
  - ...
```

I would like to have "core" module with some extra logic later on for email, passwordless, etc auth as well.

Sources/Auth/Apple/AppleAuthProvider.swift
```diff
@@ -0,0 +1,173 @@
...
+
+// MARK: - Actions
+private extension AppleAuthProvider {
+  @objc func appleCredentialRevoked() {
+    delegate?.appleAuthProviderIsAuthorized(false)
+  }
+}
+
+// MARK: - Private String Extension Methods
+private extension String {
```
markom-povio, line 166: Maybe we can move this to *Auth Core* module or to the PovioKit *Extensions* since it is not directly related only to the AppleAuth?

markom-povio (approved):
Nice addition 👍

Sources/Auth/Apple/AppleAuthProvider.swift
borut-t: I wanted to make Auth modules independent from any other. Good point though. I'll think about it.

Sources/Auth/Google/GoogleAuthProvider.swift
```diff
@@ -0,0 +1,102 @@
...
+                .compactMap { $0 }
+                .joined(separator: " ")
+              let response = AuthProvider.Response(token: auth.accessToken,
+                                                   name: fullName,
+                                                   email: userProfile?.email)
+              self?.delegate?.googleAuthProviderDidSignIn(with: response)
+            case (_, .some(let error)):
+              self?.delegate?.googleAuthProviderDidFail(with: .system(error))
+            default:
+              return
```
gentique, line 73: Should we call `self?.delegate?.googleAuthProviderDidFail(with: .unhandledAuthorization)` here?

Sources/Auth/Google/GoogleAuthProvider.swift
```diff
@@ -0,0 +1,102 @@
...
+  }
+  
+  /// Clears the signIn footprint and logs out the user immediatelly.
+  public static func signOut() {
+    GIDSignIn.sharedInstance.signOut()
+  }
+  
+  /// Checks the current auth state and returns the boolean value.
+  public static func checkAuthState(_ state: @escaping (Bool) -> Swift.Void) {
+    state(GIDSignIn.sharedInstance.currentUser != nil)
```
gentique, line 93: Is this actually `@escaping`?

Sources/Auth/Google/GoogleAuthProvider.swift
```diff
@@ -0,0 +1,102 @@
...
+              self?.delegate?.googleAuthProviderDidSignIn(with: response)
+            case (_, .some(let error)):
+              self?.delegate?.googleAuthProviderDidFail(with: .system(error))
+            default:
+              return
+            }
+          }
+        case (_, let actualError?):
+          let errorCode = (actualError as NSError).code
+          guard errorCode != GIDSignInError.Code.canceled.rawValue else { return }
```
gentique, line 78: Should it be up to the client to decide what to do with a `.canceled` error?

Sources/Auth/Google/GoogleAuthProvider.swift
borut-t: Absolutely. Looks like I've missed it.

Sources/Auth/Google/GoogleAuthProvider.swift
borut-t: It is, but only for Apple provider since method `getCredentialState` is escaping.

Sources/Auth/Google/GoogleAuthProvider.swift
borut-t: Good point. I think we need to distinct between cancelled event and errors. Thank for noticing.

gentique (approved):

Sources/Auth/Google/GoogleAuthProvider.swift
gentique: Do we have plans for adopting more of swift concurrency instead of closures?

borut-t: Not atm. You are welcome to extend it 💪

tonikocjan: We wan still return a Promise instead of using a closure. 💁🏼

Sources/Auth/Core/AuthProvider.swift
```diff
@@ -0,0 +1,38 @@
...
+//  AuthProvider.swift
+//  PovioKit
+//
+//  Created by Borut Tomazin on 25/11/2022.
+//  Copyright © 2022 Povio Inc. All rights reserved.
+//
+
+import Foundation
+
+public protocol AuthProvidable {
```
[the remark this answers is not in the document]
borut-t: > A suggestion for a different naming convention. How about instead of 'AuthProvider' we name it 'Authenticator'? Hence, 'AppleAuthenticator' 'FacebookAuthenticator', ...

I kind a like it. I have other in mind, *.Authorizer. Let's do the voting....
> 
> 
> Wouldn't the API be much nicer if the methods are async / return a Promise, instead of delegate pattern?

Done
> 
> 
> I'm not conviced if this protocol is useful or not. The issue is that each specific auth provider can have different parameters for its signing process. This is obvious by looking into Apple provider, which accepts a 'nonce' parameter and Facebook provider, which requires login permissions. Because this protocol is very rigid, the permissions in Facebook provider have to be hardcoded in the class itself (or modifiable through config). Future providers might have more specific parameters still.
> 
> I propose we get rid of this protocol and allow each of the providers to provide its own, clean, API.

I had a general solution in mind to have a central "manager" that will handle all kind of auths. But as soon as you get to a point where you need to isolate packages due to heavy dependencies, it all falls apart.

borut-t:
@poviolabs/ios-developers let's do the voting about the naming convention for the Auth* providers.

Here are the three proposed naming variants:
1️⃣ {provider}AuthProvider, e.g. `AppleAuthProvider`
2️⃣ {provider}Authenticator, e.g. `AppleAuthenticator`
3️⃣ {provider}Authorizer, e.g. `AppleAuthorizer`

React with emoji number as above...

PR #221, opened 2022-11-28
Fix/Test extensions

borut-t:
Unfortunately, we need to remove `XCTest` extension methods because they can only be included in the test targets.

We have two options to overcome this [limitation](https://forums.swift.org/t/the-error-cannot-find-auto-link-library-xctest-and-xctestswiftsupport/34785):
1. add build setting [ENABLE_TESTING_SEARCH_PATHS](https://xcodebuildsettings.com/#enable_testing_search_paths) to the project build settings that include PovioKit
2. add `XCTest.framework` manually

Both options are non-optimal since they include manual work after PovioKit is integrated into the project.

I'll move the removed extensions to the `resources` repo instead.


gentique (approved):
Is it not possible to add a product like we do for UI, Networking, named Tests that should be added to test targets only?

Or does that still produce the same effect?

borut-t:
> Is it not possible to add a product like we do for UI, Networking, named Tests that should be added to test targets only?
> 
> Or does that still produce the same effect?

Correct. `XCTest` framework is only meant for test targets. That's why.

markom-povio (approved):

PR #224, opened 2022-12-12
Fix/ANC localization error

borut-t:
An update that allows us to handle network errors when there is no data available. This becomes helpful in situations when the client returned an error without any data, e.g. connectivity or timeout errors.

Error enum was replaced by `LocalizedError` so we can extract localized description out of it, particularly from the handled errors.

### DIFF
--- a/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
+++ b/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
@@ -193,9 +193,18 @@ public extension AlamofireNetworkClient {
 
 // MARK: - Models
 public extension AlamofireNetworkClient {
-  enum Error: Swift.Error {
+  enum Error: LocalizedError {
     case request(RequestError, ErrorInfo)
     case other(Swift.Error, ErrorInfo)
+    
+    public var errorDescription: String? {
+      switch self {
+      case .request(let err, _):
+        return err.localizedDescription
+      case .other(let err, _):
+        return err.localizedDescription
+      }
+    }
   }
   
   class Request {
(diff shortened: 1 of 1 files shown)


gentique (approved):

markom-povio (approved):

tonikocjan (approved):

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -193,9 +193,18 @@ public extension AlamofireNetworkClient {
...
     case other(Swift.Error, ErrorInfo)
+    
+    public var errorDescription: String? {
+      switch self {
+      case .request(let err, _):
+        return err.localizedDescription
+      case .other(let err, _):
+        return err.localizedDescription
+      }
+    }
```
tonikocjan, line 207: Maybe we can include `ErrorInfo` into the localized description?

borut-t: The idea is that `errorDescription` returns localized description from a concrete error that was handled by the caller. Here is an example...

```swift
AlamofireNetworkClient(session: session,
                       eventMonitors: [ANCSentryErrorMonitor()]) { origError, _ in
  CustomError(
    code: 1,
    message: "My custom error message",
    detail: origError.localizedDescription
  )
}
```

This is an implementation with the default error handler in place. So each error emitted will be transformed into CustomError. You see where I'm going...

When you get error from the API, you can now just write `error.localizedDescription` and it will return "My custom error message".

You can always cast this "Error" to AlamofireNetworkClient.Error to get out `ErrorInfo` or other info as well.

PR #225, opened 2022-12-22
Chore/ANC - Doc update

borut-t:


nagavcindriqim (approved):

Resources/Networking/AlamofireNetworkClient/README.md
```diff
@@ -1,6 +1,6 @@
-# AlamofireRestClient
+# Alamofire Network Client
 
-High-level `Alamofire` REST client abstraction.
+High-level network client abstraction based on [Alamofire](https://github.com/Alamofire/Alamofire).
```
gentique, line 3: Just wondering if we can call ANC an abstraction, since moving on to let's say URLSession would be quite difficult once ANC has been implemented on a project.

borut-t: In a way, you are right since there is no pure abstraction there. But changing the underlying engine wouldn't be that hard to do. There are some Alamofire types defined for params, but that could be easily overcome.

gentique (approved):

PR #227, opened 2022-12-23
Fix/Auth providers

borut-t:
I've tested out all three providers on the real project and came up with some updates and modifications. Created a separate PR so we know exactly what was changed.

### DIFF
--- a/Sources/Auth/Apple/AppleAuthProvider.swift
+++ b/Sources/Auth/Apple/AppleAuthProvider.swift
@@ -7,12 +7,11 @@
 //
 
 import AuthenticationServices
-import CryptoKit
 import Foundation
 import PovioKitAuthCore
 
 public protocol AppleAuthProvidable: AuthProvidable {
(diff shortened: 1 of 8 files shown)


Sources/Auth/Apple/AppleAuthProvider.swift
```diff
@@ -108,16 +103,16 @@ extension AppleAuthProvider: ASAuthorizationControllerDelegate {
...
-
 // MARK: - Private Methods
 private extension AppleAuthProvider {
-  func appleSignIn(with nonce: Nonce?) {
+//  var checkedDelegate: AppleAuthProviderDelegate? {
+//    if delegate == nil {
+//      delegate?.appleAuthProviderDidFail(with: .cancelled)
+//    }
+//    return delegate
+//  }
```
nagavcindriqim, line 113: can we remove this?

borut-t: Yes sir 🖖

nagavcindriqim (approved):

PR #229, opened 2022-12-26
Feature/Auth providers Promise

borut-t:
All public APIs now support promises, replacing closures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,21 +23,24 @@
 
 ## Packages
 
-| [Core](Resources/Core) | [UI](Resources/UI) | [Networking](Resources/Networking) | [PromiseKit](Resources/PromiseKit) | [Auth](Resources)
-| :-: | :-: | :-: | :-: |
+| [Core](Resources/Core) | [UI](Resources/UI) | [Networking](Resources/Networking) | [PromiseKit](Resources/PromiseKit) | [Auth](Resources/Auth)
+| :-: | :-: | :-: | :-: | :-: |
 
 ## Installation
 
 ### Swift Package Manager
 - In Xcode, click `File` -> `Add Packages...`  
-- Insert `https://github.com/poviolabs/PovioKit` in the search field.
-- Select `Dependency Rule` "Up to Next Major Version" with "2.0.0"
-
-Currently, there are three packages to choose from:
-- *PovioKit* (core)
-- *PovioKitNetworking* (networking library, depends on `core` and `promise` package)
-- *PovioKitPromise* (lightweight promises library)
-- *PovioKitUI* (UI components)
+- Insert `https://github.com/poviolabs/PovioKit` in the Search field.
+- Select a desired `Dependency Rule`. Usually "Up to Next Major Version" with "2.0.0".
+- Select "Add Package" button and check one or all given products from the list:
+  - *PovioKit* (core)
+  - *PovioKitNetworking* (networking library, depends on `core` and `promise` package)
+  - *PovioKitPromise* (lightweight promises library)
+  - *PovioKitUI* (UI components)
+  - *PovioKitAuthApple* (Apple auth components)
+  - *PovioKitGoogleApple* (Google auth components)
+  - *PovioKitFacebookApple* (Facebook auth components)
+- Select "Add Package" again and you are done.
 
 ### Migration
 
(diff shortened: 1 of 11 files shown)


markom-povio (approved):
👍

README.md
```diff
@@ -23,21 +23,24 @@
...
-- Insert `https://github.com/poviolabs/PovioKit` in the search field.
-- Select `Dependency Rule` "Up to Next Major Version" with "2.0.0"
-
-Currently, there are three packages to choose from:
-- *PovioKit* (core)
-- *PovioKitNetworking* (networking library, depends on `core` and `promise` package)
-- *PovioKitPromise* (lightweight promises library)
-- *PovioKitUI* (UI components)
+- Insert `https://github.com/poviolabs/PovioKit` in the Search field.
+- Select a desired `Dependency Rule`. Usually "Up to Next Major Version" with "2.0.0".
```
markom-povio, line 34: Should this be "2.1.0"?

borut-t: Not necessarily. Anything equal to `2.0.0` or above will update up to version 3.0.0 anyway.

Sources/Auth/Facebook/FacebookAuthProvider.swift
```diff
@@ -65,16 +53,38 @@ extension FacebookAuthProvider: FacebookAuthProvidable {
   }
   
   /// Checks the current auth state and returns the boolean value.
-  public static func checkAuthState(_ state: @escaping (Bool) -> Swift.Void) {
+  public static func isAuthorized() -> Authorized {
     let exists = AccessToken.current?.tokenString != nil
     let isValid = !(AccessToken.current?.isExpired ?? true)
-    state(exists && isValid)
+    return exists && isValid
```
tonikocjan, line 59:
```swift
guard let token = AccessToken.current else { return false }
return !token.isExpired
```

Sources/Auth/Google/GoogleAuthProvider.swift
```diff
@@ -9,73 +9,72 @@
...
+import PovioKitPromise
 
-public protocol GoogleAuthProvidable: AuthProvidable {
+public protocol GoogleAuthProvidable {
+  typealias Authorized = Bool
+  typealias Response = AuthProvider.Response
+  
+  func signIn(from presentingViewController: UIViewController) -> Promise<Response>
+  static func signOut()
+  static func isAuthorized() -> Authorized
```
tonikocjan, line 20: Same here.

tonikocjan (changes requested):
Nice. Left a couple of suggestions.

Sources/Auth/Apple/AppleAuthProvider.swift
```diff
@@ -9,22 +9,24 @@
...
-  func appleAuthProviderCredentialsRevoked()
+public protocol AppleAuthProvidable {
+  typealias Authorized = Bool
+  typealias Response = AuthProvider.Response
+  
+  func signIn(from presentingViewController: UIViewController) -> Promise<Response>
+  func signIn(from presentingViewController: UIViewController,
+              with nonce: AppleAuthProvider.Nonce) -> Promise<Response>
+  static func signOut()
+  static func checkAuthState() -> Promise<Authorized>
```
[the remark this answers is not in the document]
borut-t: The thing about static method is that this functionality is only limited to certain scopes. E.g. to check ad-hoc if user is logged in. Initially, the provider initialization included some crucial dependencies as well, which made such operations troublesome. Now that we have a lean initialization, we could potentially go with this. 
I'll think about that if that makes sense, I'll open a separate PR.

Sources/Auth/Facebook/FacebookAuthProvider.swift
```diff
@@ -10,53 +10,41 @@ import Foundation
...
 
 // MARK: - Public Methods
 extension FacebookAuthProvider: FacebookAuthProvidable {
   /// SignIn user.
   ///
-  /// Will notify the delegate with the `Response` object on success or with `Error` on error.
-  public func signIn(on presentingViewController: UIViewController) {
+  /// Will return promise with the `Response` object on success or with `Error` on error.
+  public func signIn(from presentingViewController: UIViewController) -> Promise<Response> {
     let permissions: [String] = (defaultPermissions + config.extraPermissions).map { $0.name }
```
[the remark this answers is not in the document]
borut-t: Good point. I need to rethink this logic. A lot of changes were made, and not everything was considered. Again, topic for a separate PR.

PR #233, opened 2023-01-06
Ext/SafeAreaInsets

borut-t:
A useful extension to retrieve insets of the device for the key window.

### DIFF
--- a/Sources/Core/Extensions/UIKit/UIWindow+PovioKit.swift
+++ b/Sources/Core/Extensions/UIKit/UIWindow+PovioKit.swift
@@ -1,0 +1,18 @@
+//
+//  UIWindow+PovioKit.swift
+//  PovioKit
+//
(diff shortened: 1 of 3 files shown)


Sources/Core/Extensions/UIKit/UIWindow+PovioKit.swift
```diff
@@ -0,0 +1,18 @@
...
+//  UIWindow+PovioKit.swift
+//  PovioKit
+//
+//  Created by Borut Tomazin on 5/1/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import UIKit
+
+extension UIWindow {
```
tonikocjan, line 11: Needs to be public.

tonikocjan (approved):

markom-povio (approved):

PR #236, opened 2023-01-11
Chore/Apple Auth email metadata

borut-t:
Added `JWTDecoder` util to decode JWT tokens.

### DIFF
--- a/Sources/Auth/Apple/AppleAuthProvider.swift
+++ b/Sources/Auth/Apple/AppleAuthProvider.swift
@@ -96,7 +96,7 @@ extension AppleAuthProvider: ASAuthorizationControllerDelegate {
       // store userId for later
       Self.storage.set(credential.user, forKey: Self.userIdStorageKey)
       
-      // parse email and it's metadata
+      // parse email and related metadata
       let email: AuthProvider.Response.Email? = credential.email.map {
         let identity = try? JWTDecoder(token: identityTokenString)
         let isEmailPrivate = identity?.bool(for: "is_private_email")
--- a/Sources/Auth/Apple/AppleAuthProvider.swift
+++ b/Sources/Auth/Apple/AppleAuthProvider.swift
@@ -96,9 +96,21 @@ extension AppleAuthProvider: ASAuthorizationControllerDelegate {
       // store userId for later
       Self.storage.set(credential.user, forKey: Self.userIdStorageKey)
       
+      // parse email and it's metadata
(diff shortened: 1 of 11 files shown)


Sources/Auth/Core/JWTDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+//
+//  Created by Borut Tomazin on 10/1/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import Foundation
+
+/// JWTDecoder for decoding JSON Web Tokens (JWT) tokens
+/// Inspired by https://github.com/auth0/jwt-decode
+public struct JWTDecoder {
```
gentique, line 13: We could write some tests for this implementation.

borut-t: Didn't get there yet. A separate PR is incoming.

Sources/Auth/Apple/AppleAuthProvider.swift
```diff
@@ -96,9 +96,21 @@ extension AppleAuthProvider: ASAuthorizationControllerDelegate {
       // store userId for later
       Self.storage.set(credential.user, forKey: Self.userIdStorageKey)
       
+      // parse email and it's metadata
```
tonikocjan, line 99: These and the next two comments feel a bit unnecessary.

Sources/Auth/Core/AuthProvider.swift
```diff
@@ -31,3 +31,17 @@ public struct AuthProvider {
     case credentialsRevoked
   }
 }
+
+public extension AuthProvider.Response {
+  struct Email {
+    let address: String
+    let isPrivate: Bool?
+    let isVerified: Bool?
```
tonikocjan, line 39: Are these properties supposed to be visible to the outside world? If yes, they need to be `public`.

tonikocjan (approved):

Sources/Auth/Apple/AppleAuthProvider.swift
borut-t: I think better more than less 🤷‍♂️

Sources/Auth/Core/AuthProvider.swift
borut-t: Visible yes. Pushed an [update](https://github.com/poviolabs/PovioKit/pull/236/commits/9b3b2f960c6525ba50db2a63f69caac620a7df76).

gentique (approved):

PR #240, opened 2023-01-20
Chore/Auth updates

borut-t:
I've made some improvements and changes based on the comments from the #229.

I've also re-introduced a protocol `AuthProvidable` to which all authenticators adhere. Why? Read further...

Imagine, you have some sort of a manager that handles auth. And you are covering e.g. "Apple" and "Google" auth. User signs in with Apple. When we need to perform some action, we just call `AuthProvidable` instance and call that method. We don't really care if authenticator is Apple or Google or anything else. If we went with the previous logic, we would first need to check which provider was selected, pick that one and call a method.

What do you think?

### DIFF
--- a/Sources/Auth/Apple/AppleAuthenticator.swift
+++ b/Sources/Auth/Apple/AppleAuthenticator.swift
@@ -18,18 +18,18 @@ public protocol AppleAuthProvidable {
   func signIn(from presentingViewController: UIViewController) -> Promise<Response>
   func signIn(from presentingViewController: UIViewController,
               with nonce: AppleAuthenticator.Nonce) -> Promise<Response>
-  static func signOut()
-  static func checkAuthState() -> Promise<Authorized>
+  func signOut()
+  func checkAuthState() -> Promise<Authorized>
 }
 
 public final class AppleAuthenticator: NSObject {
-  private static let userIdStorageKey = "povioKit.appleSocialProvider.signIn.userId"
-  private static let storage: UserDefaults = .standard
-  private let authProvider: ASAuthorizationAppleIDProvider
+  private let userIdStorageKey = "povioKit.appleSocialProvider.signIn.userId"
+  private let storage: UserDefaults = .standard
+  private let provider: ASAuthorizationAppleIDProvider
   private var processingPromise: Promise<Response>?
   
   public override init() {
-    self.authProvider = .init()
+    self.provider = .init()
     super.init()
     setupCredentialsRevokeListener()
   }
@@ -63,13 +63,13 @@ extension AppleAuthenticator: AppleAuthProvidable {
   }
   
   /// Clears the signIn footprint and logs out the user immediatelly.
-  public static func signOut() {
+  public func signOut() {
     storage.removeObject(forKey: userIdStorageKey)
   }
   
   /// Checks the current auth state and returns the boolean value as promise.
-  public static func checkAuthState() -> PovioKitPromise.Promise<Authorized> {
-    guard let userId = Self.storage.string(forKey: Self.userIdStorageKey) else {
+  public func checkAuthState() -> Promise<Authorized> {
+    guard let userId = storage.string(forKey: userIdStorageKey) else {
       return .value(false)
     }
     
@@ -93,7 +93,7 @@ extension AppleAuthenticator: ASAuthorizationControllerDelegate {
       }
       
       // store userId for later
-      Self.storage.set(credential.user, forKey: Self.userIdStorageKey)
+      storage.set(credential.user, forKey: userIdStorageKey)
       
       // parse email and related metadata
       let email: AuthProvider.Response.Email? = credential.email.map {
@@ -125,7 +125,7 @@ extension AppleAuthenticator: ASAuthorizationControllerDelegate {
 // MARK: - Private Methods
 private extension AppleAuthenticator {
   func appleSignIn(on presentingViewController: UIViewController, with nonce: Nonce?) {
-    let request = authProvider.createRequest()
+    let request = provider.createRequest()
     request.requestedScopes = [.fullName, .email]
     
     switch nonce {
--- a/Sources/Auth/Apple/AppleAuthenticator.swift
+++ b/Sources/Auth/Apple/AppleAuthenticator.swift
@@ -59,6 +59,8 @@ extension AppleAuthenticator: AppleAuthProvidable {
   
   /// Clears the signIn footprint and logs out the user immediatelly.
   public func signOut() {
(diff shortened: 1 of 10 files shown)


Sources/Auth/Apple/AppleAuthenticator.swift
```diff
@@ -63,13 +58,13 @@ extension AppleAuthenticator: AppleAuthProvidable {
   }
   
   /// Clears the signIn footprint and logs out the user immediatelly.
-  public static func signOut() {
+  public func signOut() {
     storage.removeObject(forKey: userIdStorageKey)
```
markom-povio, line 62: Should we reject current processingPromise also?

markom-povio (approved):

Sources/Auth/Apple/AppleAuthenticator.swift
borut-t: Good point 👍 [Here](https://github.com/poviolabs/PovioKit/pull/240/commits/f29f82f066d3a2348fe9a076ae7ae39f69a42caa) it is...

Sources/Auth/Core/AuthProvidable.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+
+import PovioKitPromise
+import UIKit
+
+public protocol AuthProvidable {
+  typealias Authenticated = Bool
+  typealias Response = Authenticator.Response
+  
+  func signIn(from presentingViewController: UIViewController) -> Promise<Response>
```
tonikocjan, line 16: I think we can use [primary associated types](https://www.donnywals.com/what-are-primary-associated-types-in-swift-5-7/) to improve the design a bit.

We know that `Authenticator`s require `permissions` when logging in. Hence, it would make sense to add an associated type `Permission` to the protocol and change `login` function to accept a list of permissions.
But we know that this would lead to the old _protocol 'AuthProvidable can only be used as a generic constraint ..._ error.

Using primary associated types we can, I believe, avoid this issue. Here is how it would look in action:

```swift
public protocol AuthProvidable<Permission> {
    associatedtype Permission

    func signIn(
      permissions: [Permission]
      from presentingViewController: UIViewController
    ) -> Promise<Response>

    func signIn(
      from presentingViewController: UIViewController
    ) -> Promise<Response>
}

extension AuthProvidable {
    // implementations can provide their own methods, 
    // which can use _default permissions_
    func signIn(
      from presentingViewController: UIViewController
    ) -> Promise<Response> {
      signIn(permissions: [], from: presentingViewController)
    }
}
```

Specific protocols would look like this:

```swift
public protocol FacebookAuthProvidable: AuthProvidable<Permission> {
  var isAuthenticated: Authenticated { get }
}

public protocol GoogleAuthProvidable: AuthProvidable<()> {
  var isAuthenticated: Authenticated { get }
  static func shouldHandleURL(_ url: URL) -> Bool
}
```

---

What do you think?

Sources/Auth/Facebook/FacebookAuthenticator.swift
```diff
@@ -12,58 +12,65 @@ import PovioKit
...
-  typealias Authorized = Bool
-  typealias Response = AuthProvider.Response
-  
-  func signIn(from presentingViewController: UIViewController) -> Promise<Response>
-  static func signOut()
-  static func isAuthorized() -> Authorized
+public protocol FacebookAuthProvidable: AuthProvidable {
+  func signIn(from presentingViewController: UIViewController,
+              with extraPermissions: [Permission]) -> Promise<Response>
+  var isAuthenticated: Authenticated { get }
```
tonikocjan, line 18: Does this and other such protocols add any value? For instance, shouldn't `var isAuthenticated: Authenticated { get }` be part of all `Authenticator`s?

Sources/Auth/Core/AuthProvidable.swift
```diff
@@ -0,0 +1,18 @@
...
+//  PovioKit
+//
+//  Created by Borut Tomazin on 20/01/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import PovioKitPromise
+import UIKit
+
+public protocol AuthProvidable {
```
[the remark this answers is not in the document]
borut-t: I was already on the verge of doing that, but smth didn't add up. I'll check again.

Sources/Auth/Core/AuthProvidable.swift
borut-t: That's all good until it isn't. Every provider has it's own permissions to handle. This generic solution could solve this, but we'll leave that logic to the user to figure out the exact types for each platform. Which for me is not ideal.
I'll think about it

Sources/Auth/Facebook/FacebookAuthenticator.swift
borut-t: The thing is, all but AppleAuthenticator has a state for it. Only Apple requires to query it. Therefore, we cannot use the same getter for all instances.

Sources/Auth/Core/AuthProvidable.swift
tonikocjan: Exactly. Each provider has its own permissions, which means that when you want to use a specific provider, you accept `any` implementation of `AuthProvidable`, which knows how to work with such permissions; i.e.:

```swift
func signInToFacebook(authenticator: any AuthProvidable<FacebookPermission>) -> ..
```

Sources/Auth/Facebook/FacebookAuthenticator.swift
tonikocjan: How about converting it into a `Promise`? Those that don't need to query can return `.value(state)`, whereas for Apple's you do the querying ..

borut-t: Not perfect. How would a dev know where to pick which?

tonikocjan: What I mean is that there is only one getter.

```swift
protocol AuthProvidable {
  ...
  var authenticated: Promise<Bool> { get }
}
```

borut-t: I understand. But not optimal. It's not optimal in any way either. I had this on initial implementation but abandoned it later. I'll revisit.

borut-t:
@tonikocjan I'll merge this PR and #245 and make changes on a separate PR.

Sources/Auth/Facebook/FacebookAuthenticator.swift
tonikocjan: Just a suggestion. :)

PR #246, opened 2023-01-25
Chore/SPM package update checker

borut-t:
Periodically check for outdated dependencies. Currently no action is taken, but we could send a slack message when outdated packages are found.


markom-povio (approved):
Nice idea 👍

tonikocjan (approved):

PR #248, opened 2023-01-25
Feature: InAppPurchase

markom-povio:
In-app purchase library done with async/await, Result type, and StoreKit 2 implementation. Minimum deployment targets are from iOS 15.0 and up. Functionality:
- Purchase plan with options
- Check if the plan is purchased
- Restore purchases
- Validate receipt

Example:
``` swift
Task {
  let inAppPurchase = InAppPurchase(identifiers:  ["com.test.plan1", "com.test.plan2", "com.test.plan3", "com.test.plan4"])
  let result = await inAppPurchase.purchase(plan: "com.test.plan1")
  switch result {
    case .success(let transaction):
      print("purchase plan success with transaction: \(transaction.productID)")
    case .failure(let error):
      print("purchase plan failure with error: \(error.localizedDescription)")
  }
}
```


Sources/Core/Utilities/InAppPurchase/InAppPurchase.swift
```diff
@@ -0,0 +1,197 @@
...
+  /// - Parameters:
+  ///   - plan: InAppPurchase plan identifier (eg. "com.test.plan") to purchase
+  ///   - options: Set of ``PurchaseOption``
+  /// - Returns: Result type with completed ``Transaction`` object or ``InAppPurchaseError``
+  /// - Important: ``InAppPurchaseError`` cases that this function returns:
+  /// * ``InAppPurchaseError.missingProductId``
+  /// * ``InAppPurchaseError.paymentCancelled``
+  /// * ``InAppPurchaseError.paymentPending``
+  /// * ``InAppPurchaseError.requestFailed(error)``
+  public func purchase(plan: IAPPlan, options: Set<Product.PurchaseOption> = []) async -> Result<Transaction, InAppPurchaseError> {
```
tonikocjan, line 54: Did you consider making the function throwable, instead of returning Result? What are pros/cons of either approach?

markom-povio: Yes, I did, but I just like how the Result is handled in the implementation code, and no need for a do/catch block. It is cleaner for the user in my opinion. Do you have any thoughts on this?

tonikocjan: From what I've seen, the common approach is to use `async throws` in Apple's APIs and elsewhere. The advantage is, probably, that you don't have to handle error cases for every async expression; `do/catch` does it for you. But I don't have a strong preference at this point. 🤷

ArlindDushi (approved):
Great work! 👌

nagavcindriqim (approved):
🙌

gentique:
Hmm `InAppPurchase` sounds somehow like a type to me, would a name like `InAppPurchaseService` make sense? 

Also the in-app SKU-s is typealiased as `InAppPlan` but internally is referred as product, so wondering if `InAppProduct` would be more appropriate, since plan generally refers to something broader than a singular "thing".

Thoughts?

gentique:
Hmm `InAppPurchase` sounds somehow like a type to me, would a name like `InAppPurchaseService` make sense? 

Also the in-app SKU-s is typealiased as `InAppPlan` but internally is referred as product, so wondering if `InAppProduct` would be more appropriate, since plan generally refers to something broader than a singular "thing".

Thoughts?

markom-povio:
> Hmm `InAppPurchase` sounds somehow like a type to me, would a name like `InAppPurchaseService` make sense?
> 
> Also the in-app SKU-s is typealiased as `InAppPlan` but internally is referred as product, so wondering if `InAppProduct` would be more appropriate, since plan generally refers to something broader than a singular "thing".
> 
> Thoughts?

I agree, it makes sense.

borut-t (approved):
Awesome 👌

PR #249, opened 2023-01-25
ANC/Requests with header

tonikocjan:
Had a use-case when integrating the new LinkedIn API that some information had to be extracted from the response header. This PR extends ANC to support this.

### DIFF
--- a/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
+++ b/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
@@ -253,35 +253,13 @@ public extension AlamofireNetworkClient.Error {
 // MARK: - Request API
 public extension AlamofireNetworkClient.Request {
   var asData: Promise<Data> {
-    .init { promise in
-      dataRequest.responseData { (response: AFDataResponse<Data>) in
-        switch response.result {
-        case .success(let data):
-          self.eventMonitors.forEach { $0.requestDidSucceed(self) }
-          promise.resolve(with: data)
-        case .failure(let error):
-          let error = self.handleError(error)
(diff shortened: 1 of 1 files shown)


markom-povio (approved):

ArlindDushi (approved):
Looks good 👍

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -253,12 +253,31 @@ public extension AlamofireNetworkClient.Error {
...
+  func decode<D: Decodable>(
+    _ decodable: D.Type,
+    decoder: JSONDecoder = .init(),
+    on dispatchQueue: DispatchQueue? = .main
+  ) -> Promise<D> {
+    decodeWithHeaders(decodable, decoder: decoder, on: dispatchQueue)
+      .map { $0.0 }
+  }
+  
+  var asDataWithHeaders: Promise<(Data, HTTPHeaders?)> {
```
borut-t, line 274: Instead of introducing new `*withHeaders` methods, should we rather always return headers instead?
e.g.
```swift
var asData: Promise<(Data, Headers?)>
var asVoid: Promise<(Headers?)>
```

tonikocjan: This would break all the existing code.

borut-t: You are right. Since this is really and edge case, we will probably never use it. And to tie it to existing code, doesn't make much sense.

borut-t (approved):

Issue #241, opened 2023-01-23
[UI] ActionButton refinements


borut-t:
The current state of the ActionButton is great, but could be improved. Here is a few things that we could work on:
- extract `properties` directly to view modifiers instead
- build up all [available](https://www.figma.com/file/75s6aqhuVmZbUKnNTXfOKF/Povio-UX-Library---Mobile?node-id=892%3A52717&t=bIvdGXEodBUl1YR9-0) customizations

PR #250, opened 2023-01-27
UI/ActionButton refinements

ArlindDushi:
Action button refinements to:
- Make access/changing of properties easier in UIKit & SwiftUI
- Cover all customizations in Povio UX Mobile Library
Screenshot 2023-01-27 at 10 08 21 AM

### DIFF
--- a/Sources/UI/ActionButton/ActionButton.swift
+++ b/Sources/UI/ActionButton/ActionButton.swift
@@ -72,18 +72,18 @@ public struct ActionButton: View {
 // MARK: - ActionButtonViewModel
 private extension ActionButton {
   class ActionButtonViewModel: ObservableObject {
-    @Published var title: String = "Button"
-    @Published var font: Font = .system(size: 16)
-    @Published var textColor: Color = .white
-    @Published var cornerRadius: ActionButton.CornerRadiusType = .rounded
-    @Published var backgroundType = ActionButton.Background.plain(.blue)
-    @Published var borderColor: Color = .clear
-    @Published var borderWidth: CGFloat = 0
-    @Published var leftImage: ActionButton.ExtraImage?
-    @Published var rightImage: ActionButton.ExtraImage?
-    @Published var titleLeftImage: ActionButton.ExtraImage?
-    @Published var titleRightImage: ActionButton.ExtraImage?
-    @Published var actionHandler: (() -> Void)?
+    @Published public var title: String = "Button"
+    @Published public var font: Font = .system(size: 16)
+    @Published public var textColor: Color = .white
+    @Published public var cornerRadius: ActionButton.CornerRadiusType = .rounded
+    @Published public var backgroundType = ActionButton.Background.plain(.blue)
+    @Published public var borderColor: Color = .clear
+    @Published public var borderWidth: CGFloat = 0
+    @Published public var leftImage: ActionButton.ExtraImage?
+    @Published public var rightImage: ActionButton.ExtraImage?
+    @Published public var titleLeftImage: ActionButton.ExtraImage?
+    @Published public var titleRightImage: ActionButton.ExtraImage?
+    @Published public var actionHandler: (() -> Void)?
   }
 }
 
--- a/Sources/UI/ActionButton/ActionButton.swift
+++ b/Sources/UI/ActionButton/ActionButton.swift
@@ -72,18 +72,18 @@ public struct ActionButton: View {
 // MARK: - ActionButtonViewModel
 private extension ActionButton {
   class ActionButtonViewModel: ObservableObject {
-    @Published public var title: String = "Button"
-    @Published public var font: Font = .system(size: 16)
-    @Published public var textColor: Color = .white
-    @Published public var cornerRadius: ActionButton.CornerRadiusType = .rounded
-    @Published public var backgroundType = ActionButton.Background.plain(.blue)
-    @Published public var borderColor: Color = .clear
-    @Published public var borderWidth: CGFloat = 0
-    @Published public var leftImage: ActionButton.ExtraImage?
-    @Published public var rightImage: ActionButton.ExtraImage?
-    @Published public var titleLeftImage: ActionButton.ExtraImage?
-    @Published public var titleRightImage: ActionButton.ExtraImage?
-    @Published public var actionHandler: (() -> Void)?
+    @Published var title: String = "Button"
+    @Published var font: Font = .system(size: 16)
+    @Published var textColor: Color = .white
+    @Published var cornerRadius: ActionButton.CornerRadiusType = .rounded
+    @Published var backgroundType = ActionButton.Background.plain(.blue)
+    @Published var borderColor: Color = .clear
+    @Published var borderWidth: CGFloat = 0
+    @Published var leftImage: ActionButton.ExtraImage?
+    @Published var rightImage: ActionButton.ExtraImage?
+    @Published var titleLeftImage: ActionButton.ExtraImage?
+    @Published var titleRightImage: ActionButton.ExtraImage?
+    @Published var actionHandler: (() -> Void)?
   }
 }
 
--- a/Sources/UI/ActionButton/ActionButton.swift
+++ b/Sources/UI/ActionButton/ActionButton.swift
@@ -8,7 +8,7 @@
 
 import SwiftUI
 
-public class ActionButtonViewModel: ObservableObject {
+private class ActionButtonViewModel: ObservableObject {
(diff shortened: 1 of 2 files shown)


ArlindDushi:
Closes #241

ArlindDushi:
resolve #241

ArlindDushi:
Closes #241

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -33,25 +39,56 @@ public struct ActionButton: View {
...
+                   height: properties.titleRightImage?.size.height)
+        }
+        Spacer()
+        properties.rightImage?.image
+          .resizable()
+          .aspectRatio(contentMode: .fit)
+          .frame(width: properties.rightImage?.size.width,
+                 height: properties.rightImage?.size.height)
       }
+      .padding()
```
markom-povio, line 73: Should we add a property in the ViewModel to control padding also?

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -33,25 +39,56 @@ public struct ActionButton: View {
...
+    }) {
+      HStack {
+        properties.leftImage?.image
+          .resizable()
+          .aspectRatio(contentMode: .fit)
+          .frame(width: properties.leftImage?.size.width,
+                 height: properties.leftImage?.size.height)
+        Spacer()
+        HStack {
+          properties.titleLeftImage?.image
```
markom-povio, line 53: Do we need both "leftImage" and "titleLeftImage"? 🤔

markom-povio (approved):
👍

Sources/UI/ActionButton/ActionButton.swift
ArlindDushi: The reason why I kept it this way is because default padding works really good for the customization types that we have.
According to documentation:   _If you set the value to nil, SwiftUI uses a platform-specific default amount_.
Either way I don't have any strong feelings about this.

Sources/UI/ActionButton/ActionButton.swift
ArlindDushi: `leftImage` is the one on the left almost bordering the edge and the `titleLeftImage` is the one right next to the title. 
They are in different stacks. We will have cases where both will be used maybe not simultaneously, but I didn't want to constraint to using just one at a time and left all options open.

For eg. we could have a button with a left image and then an arrow next to the button title. In this case we need both of them.
Even though the file doesn't have a button like that I wanted to leave this kind of flexibility.

borut-t: @ArlindDushi should we rename `titleLeftImage` to `titleImage`?

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -8,6 +8,7 @@
 
 import SwiftUI
 
+@available(iOS 13, *)
```
borut-t, line 11: We don't need this. PovioKit is built for iOS 13+ already.

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -60,111 +100,158 @@ public extension ActionButton {
...
-        return .center
-      }
-    case .none:
-      return .trailing
-    }
+// MARK: - Builder Pattern Methods
+@available(iOS 13, *)
+public extension ActionButton {
+  func title(_ title: String) -> ActionButton {
+    self.properties.title = title
```
borut-t, line 107: `self` keyword is redundant, right?

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -60,111 +100,158 @@ public extension ActionButton {
...
+  
+  func borderWidth(_ borderWidth: CGFloat) -> ActionButton {
+    self.properties.borderWidth = borderWidth
+    return self
+  }
+  
+  func extraImage(leftImage: ActionButton.ExtraImage) -> ActionButton {
+    self.properties.leftImage = leftImage
+    return self
+  }
```
borut-t, line 144: Why don't we simply name this method as `leftImage`?

Sources/UI/ActionButton/ActionButton.swift
ArlindDushi: Yes. Will remove it.

Sources/UI/ActionButton/ActionButton.swift
ArlindDushi: You're right, sounds better. I'll rename all 4 methods responsible for the images accordingly.

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -8,22 +8,30 @@
...
+  @Published public var cornerRadius: ActionButton.CornerRadiusType = .rounded
   @Published public var backgroundType = ActionButton.Background.plain(.blue)
   @Published public var borderColor: Color = .clear
   @Published public var borderWidth: CGFloat = 0
-  @Published public var extraImage: ActionButton.ExtraImage?
+  @Published public var leftImage: ActionButton.ExtraImage?
+  @Published public var rightImage: ActionButton.ExtraImage?
+  @Published public var titleLeftImage: ActionButton.ExtraImage?
+  @Published public var titleRightImage: ActionButton.ExtraImage?
+  @Published public var actionHandler: (() -> Void)?
```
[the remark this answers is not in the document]
ArlindDushi: This used to be public when we accessed the properties directly, now since we have made the access through builder methods we don't need it as this anymore, but forgot to move the definition inside the ActionButton.  Nice catch!

borut-t (approved):
👌

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -27,43 +27,50 @@ public struct ActionButton: View {
...
+            .padding(.trailing, 10)
         }
+        
         Spacer()
-        properties.rightImage?.image
-          .resizable()
-          .aspectRatio(contentMode: .fit)
-          .frame(width: properties.rightImage?.size.width,
-                 height: properties.rightImage?.size.height)
+          .frame(minHeight: 1)
```
borut-t, line 65: Is that ok to have 1 here?

borut-t (approved):

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -8,22 +8,13 @@
...
 
-public class ActionButtonViewModel: ObservableObject {
-  @Published public var title: String = ""
-  @Published public var font: Font = .system(size: 18)
-  @Published public var textColor: Color = .white
-  @Published public var cornerRadius: ActionButton.CornerRadiusType = .custom(0)
-  @Published public var backgroundType = ActionButton.Background.plain(.blue)
-  @Published public var borderColor: Color = .clear
-  @Published public var borderWidth: CGFloat = 0
-  @Published public var extraImage: ActionButton.ExtraImage?
```
gentique, line 19: Any particular reason why we need an `ObservableObject` for the View's properties?

Sources/UI/ActionButton/ActionButton.swift
ArlindDushi: If we go without it Spacer will use a default value which will be greater than 1. That won't allow us to go with a lower height than that default value. I don't think we'll need super-thin buttons but still I just didn't want to have that constraint.

Sources/UI/ActionButton/ActionButton.swift
gentique: Ahh, I see! Interesting. What triggered me to ask this question was the builder pattern that you used, from previous experience usually you create a copy of self, modify the intended property, then return it, whereas here it was through a `ObservableObject`. 

Thanks for explaining!

Sources/UI/ActionButton/ActionButton.swift
```diff
@@ -60,111 +112,154 @@ public extension ActionButton {
...
+    set { properties.titleLeftImage = newValue}
+  }
+  
+  var titleRightImage: ActionButton.ExtraImage? {
+    get { properties.titleRightImage }
+    set { properties.titleRightImage = newValue}
+  }
+  
+  mutating func addAction(_ action: @escaping () -> Void) {
+    properties.actionHandler = action
```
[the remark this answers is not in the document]
ArlindDushi: 1. We don't need it, legacy code that forgot to change. Nice catch! Will change it now.
2. I think I had it as `set` initially but then honestly just went with `UIButton`'s method naming convention `addAction(...)`

PR #251, opened 2023-01-31
Auth/Improvements

borut-t:
The last batch of updates.


Sources/Auth/Apple/AppleAuthenticator.swift
```diff
@@ -11,20 +11,15 @@ import Foundation
...
-  private let userIdStorageKey = "povioKit.appleSocialProvider.signIn.userId"
-  private let storage: UserDefaults = .standard
+  private let userIdStorageKey = "signIn.userId"
+  private let storage: UserDefaults
   private let provider: ASAuthorizationAppleIDProvider
   private var processingPromise: Promise<Response>?
   
   public override init() {
     self.provider = .init()
+    self.storage = .init(suiteName: "povioKit.appleAuthenticator") ?? .standard
```
tonikocjan, line 22: Should we allow the user to initialise the `storage`?

Sources/Auth/Core/SocialAuthenticator.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+//  Created by Borut Tomazin on 30/01/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import PovioKitPromise
+import Foundation
+import UIKit
+
+public class SocialAuthenticator {
```
tonikocjan, line 13: Maybe rename to smth like `SocialAuthenticationManager`?

Sources/Auth/Core/SocialAuthenticator.swift
```diff
@@ -0,0 +1,35 @@
...
+  private let authenticators: [Authenticator]
+  
+  public init(authenticators: [Authenticator]) {
+    self.authenticators = authenticators
+  }
+}
+
+extension SocialAuthenticator {
```
tonikocjan, line 21: Should we handle URLs as well?

```swift
public func canOpenUrl(_ url: URL, application: UIApplication, options: [UIApplication.OpenURLOptionsKey: Any]) -> Bool {
  for auth in authenticators {
    if auth.canOpenUrl(url, application, options) { return true }
  }
  return false
}
```

tonikocjan (approved):
This looks great!

Sources/Auth/Core/SocialAuthenticator.swift
```diff
@@ -0,0 +1,35 @@
...
+  public func authenticator<A: Authenticator>(for type: A.Type) -> A? {
+    authenticators.first { $0 is A } as? A
+  }
+  
+  public func signOut() {
```
tonikocjan, line 26: This should return a promise. Another implementation would be to just call `signOut` on all the authenticators.

Sources/Auth/Apple/AppleAuthenticator.swift
borut-t: I think we should yes.

Sources/Auth/Core/SocialAuthenticator.swift
borut-t: Why?

markom-povio (approved):
👍

PR #253, opened 2023-02-01
Chore/Modifications

borut-t:
`isAuthenticated` is now synchronous operation.

### DIFF
--- a/Sources/Auth/Apple/AppleAuthenticator.swift
+++ b/Sources/Auth/Apple/AppleAuthenticator.swift
@@ -55,10 +55,8 @@ extension AppleAuthenticator: Authenticator {
   
   /// Clears the signIn footprint and logs out the user immediatelly.
   public func signOut() {
-    processingPromise?.reject(with: Error.cancelled)
-    processingPromise = nil
     storage.removeObject(forKey: storageUserIdKey)
-    storage.removeObject(forKey: storageAuthenticatedKey)
+    rejectSignIn(with: .cancelled)
   }
   
   /// Returns the current authentication state.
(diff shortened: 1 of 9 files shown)


Sources/Auth/Apple/AppleAuthenticator.swift
```diff
@@ -166,11 +173,16 @@ private extension AppleAuthenticator {
                                            name: ASAuthorizationAppleIDProvider.credentialRevokedNotification,
                                            object: nil)
   }
+  
+  func rejectSignIn(with error: Error) {
+    storage.setValue(false, forKey: storageAuthenticatedKey)
+    processingPromise?.reject(with: error)
```
tonikocjan, line 179: `processingPromise` should probably be deleted, right?

Sources/Auth/Core/SocialAuthenticationManager.swift
```diff
@@ -27,8 +35,8 @@ extension SocialAuthenticationManager {
     authenticators.forEach { $0.signOut() }
   }
   
-  public var isAuthenticated: PovioKitPromise.Promise<Authenticator.Authenticated> {
-    all(promises: authenticators.map { $0.isAuthenticated })
-      .map { $0.contains(true) }
+  public func canOpenUrl(_ url: URL, application: UIApplication, options: [UIApplication.OpenURLOptionsKey : Any]) -> Bool {
+    guard let activeAuthenticator = authenticators.first(where: { $0.isAuthenticated }) else { return false }
```
tonikocjan, line 39: You can use the computed property instead.

tonikocjan (approved):
🙌

Sources/Auth/Core/SocialAuthenticationManager.swift
borut-t: Indeed 👍

PR #259, opened 2023-02-13
UI/Storybook

borut-t:
Created a storybook where all our UI components will reside.

Simulator Screen Shot - iPhone 14 Pro - 2023-02-13 at 15 16 25 2


Demo/Storybook/Storybook/Component.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  Component.swift
+//  Storybook
+//
+//  Created by Borut Tomazin on 23/01/2023.
+//
+
+import Foundation
+
+enum Component: String, Identifiable, Hashable, CaseIterable {
```
gentique, line 10: Based on the usage we can omit conformance to all except `CaseIterable`.

on `ContentView` we can change the ForEach method to the following 
```swift
ForEach(Component.allCases, id: \.self)
```

gentique (approved):
Except for the minor nitpick, this looks really good! I like the direction this is taking 🚀

Demo/Storybook/Storybook/Component.swift
```diff
@@ -0,0 +1,24 @@
...
+  case actionButton
+  case paddingLabel
+  
+  var id: RawValue { rawValue }
```
gentique, line 14: This can also be omitted.

Demo/Storybook/Storybook/Component.swift
borut-t: True. Thanks 🙌

markom-povio (approved):
Nice 👍

PR #262, opened 2023-02-15
ANC/Domain model transformation

tonikocjan:
A common pattern when implementing APIs is to transform API response models into domain models. Something like this:

```swift
client
  .request(method: .get, endpoint: Endpoints.getUserURL.url)
  .validate()
  .decode(UserDTO.self)
  .compactMap(with: UserMapper.transform)
```

However, I've noticed two things that are not optimal with the above code. Firstly, this kind of code gets repetitive throughout the codebase with lots of duplication. Second, even more important, the transformation process gets executed on the main thread! This might be a big issue if transforming large objects.

This PR extends ANC to support and enable this pattern out of the box. Both issues are solved at the same time, and the call site looks very similar to the above example:

```swift
client
  .request(method: .get, endpoint: Endpoints.getUserURL.url)
  .validate()
  .decode(UserDTO.self, transformToDomainModelWith: UserMapper.transform)
```

By default, the `transformToDomainModelWith` closure is executed on the `.global()` queue, but can also be configured to be executed on another one:

```swift
let mapDispatchQueue: DipatchQueue = ...
client
  .request(method: .get, endpoint: Endpoints.getUserURL)
  .validate()
  .decode(
    UserDTO.self, 
    mapToDomainModelWith: UserMapper.transform,
    mapOn: mapDispatchQueue
  )
```


borut-t (approved):
Clean solution ignoring the concerns we have.

> One concern regarding this approach is that we are mixing library code with concrete usage code, which might be suboptimal. Any thoughts?

I don't think this is a serious issue. If we would be strict, transformations to domain models should not be done on the network layer already, since this is not part of it.
I think this is a better approach than the second suggested with the promisKit, since that doesn't have anything to do with promises, therefore should not be there.

nagavcindriqim (approved):
Nice solution! 🙌

borut-t (approved):
I like this better since it's not tied to the client directly any can be used anywhere.

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -467,7 +467,7 @@ public extension RequestMonitor {
 /// promise executes its observers on the background queue as well - doing any UI
 /// related work, without dispatching to main, would crash the app!
 /// 
-/// To solve both problems, we provide a small abstraction to PromiseKit, which
+/// To solve both problems, we provide a small abstraction , which
```
borut-t, line 470: Redundant space before the comma?

borut-t (approved):

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -438,3 +438,77 @@ public extension RequestMonitor {
...
+  ) -> Promise<T> where T == Value.DomainModel {
+    compactMap(on: transformOn) {
+      try $0.toDomainModel()
+    }
+    .map(on: resolveOn) { $0 }
+  }
+}
+
+
+public extension Promise where Value: Collection, Value.Element: DomainConvertible {
```
borut-t, line 504: Should we move all these definitions into an extension of this class?

markom-povio (approved):
Nice!

Sources/Networking/AlamofireNetworkClient/DomainConvertible.swift
```diff
@@ -0,0 +1,83 @@
...
+    resolveOn: DispatchQueue = .main
+  ) -> Promise<T> where T == Value.DomainModel {
+    compactMap(on: transformOn) {
+      try $0.toDomainModel()
+    }
+    .map(on: resolveOn) { $0 }
+  }
+}
+
+
```
borut-t, line 72: Redundant new line.

borut-t (approved):

PR #264, opened 2023-02-20
Feature: [Core] EXIF reader & writer

markom-povio:
We should build a tool that will retrieve EXIF information from images and be able to write it back as well.

Should be able to:
- [x] read EXIF info from an image
- [x] write exit info to an image
- [x] make it lossless
- [x] tests

```swift
do {
    // Init
    let data = try Data(contentsOf: imageUrl)
    let manager = ExifManager(source: .data(data))
    
    // Read EXIF
    manager.readExif().finally { exif, error in
        if let exif {
            Logger.debug("Image metadata:", params: ["metadata": exif.description])
        }
    }

    // Write EXIF
    let metadata = [kCGImagePropertyExifUserComment: "This is a really nice picture."]
    manager.updateExif(metadata).finally { data, error in
        if let data {
            Logger.debug("Image metadata saved!")
        }
    }
} catch {
      Logger.error("Could not create ExifManager", params: ["reason": error.localizedDescription])
}
```


Sources/Core/Utilities/ExifManager/ExifImageSource.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+//  Created by Marko Mijatovic on 17/02/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import Foundation
+
+public enum ExifImageSource {
+  case url(URL)
+  case data(Data)
```
borut-t, line 13: Should we also accept `UIImage`?

Sources/Core/Utilities/ExifManager/ExifManager.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return .init(fulfill: imageProperties)
+  }
+    
+  /// Update EXIF metadata with the new values
+  ///
+  /// Keys for the new values must be part of the [EXIF Dictionary Keys](https://developer.apple.com/documentation/imageio/exif_dictionary_keys)
+  /// - Parameter newValue: Dictionary with the new EXIF values
+  /// - Returns: New image Data as a Promise or ``ExifError``
+  func updateExif(_ newValue: [CFString: String]) -> Promise<Data> {
```
borut-t, line 41: If we rename the manager, maybe we can rename this method as well to `update` 🤷‍♂️

borut-t (approved):
Great work 👌

Sources/Core/Utilities/ExifManager/ExifImageSource.swift
markom-povio: I can add UIImage, but to get the EXIF data, I need to convert it to the Data with `pngData()` (or `jpegData(compressionQuality)`) which will cause a loss of data in the process.

Sources/Core/Utilities/ExifManager/ExifManager.swift
```diff
@@ -0,0 +1,91 @@
...
+//  Created by Marko Mijatovic on 16/02/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import Foundation
+import ImageIO
+import PovioKitPromise
+
+/// A wrapper around Image I/O framework APIs that can be used to modify EXIF and other image metadata without recompressing the image data.
+public final class ExifManager {
```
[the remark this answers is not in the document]
markom-povio: Not sure...
We are manipulating EXIF image data, maybe using only exif as a name could be misleading.

Sources/Core/Utilities/Exif/Exif.swift
```diff
@@ -77,14 +77,18 @@ public extension ExifManager {
...
     var imageSource: CGImageSource?
     switch source {
     case .url(let url):
       imageSource = CGImageSourceCreateWithURL(url as CFURL, nil)
     case .data(let data):
       imageSource = CGImageSourceCreateWithData(data as CFData, nil)
+    case .image(let image):
+      if let data = image.pngData() {
+        imageSource = CGImageSourceCreateWithData(data as CFData, nil)
+      }
```
borut-t, line 91: Not sure if this will work. What happens, if the given `image` is a jpg image?
Let's remove this for now and we'll add support for "image" later.

borut-t (approved):
👌

Sources/Core/Utilities/Exif/ExifImageSource.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  ExifImageSource.swift
+//  PovioKit
+//
+//  Created by Marko Mijatovic on 17/02/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
```
gentique, line 10: Please only import the relevant one.

gentique (approved):

Sources/Core/Utilities/Exif/ExifImageSource.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+//  Created by Marko Mijatovic on 17/02/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import Foundation
+
+public enum ExifImageSource {
+  case url(URL)
+  case data(Data)
```
borut-t, line 13: Can we add a comment for the `data` case?
```swift
/// Keep in mind that some image formats are not lossless and could lose quality.
/// When dealing with PNGs, you could load it with `UIImage(named:` and convert it to Data with `pngData()` method. This will persist in the quality since PNG is a lossless format.
/// However, this isn't true for the JPG format which isn't lossless. You need to load Data directly from the file to avoid losing on quality with `Data(contentsOf: URL)`.
```

borut-t (changes requested):
I would like to see some tests backing this solution works as expected.

PR #266, opened 2023-02-24
UI/Profile imageview refinement

ArlindDushi:
- Make access/changing of properties easier in UIKit & SwiftUI
- Add padding, tint color and content mode to the badge image
- Update documentation

### DIFF
--- a/Sources/UI/Extensions/Image+PovioKit.swift
+++ b/Sources/UI/Extensions/Image+PovioKit.swift
@@ -8,11 +8,13 @@
 
 import SwiftUI
 
+@available(iOS 14, *)
 public extension Image {}
 
 #if canImport(Kingfisher)
 import Kingfisher
 
+@available(iOS 14, *)
 private extension Image {
   func resolveWithKingFisher(from url: URL?, placeholder: Image?) -> some View {
     KFImage(url)
--- a/Sources/UI/Extensions/Image+PovioKit.swift
+++ b/Sources/UI/Extensions/Image+PovioKit.swift
@@ -8,13 +8,11 @@
 
 import SwiftUI
 
-@available(iOS 14, *)
 public extension Image {}
 
 #if canImport(Kingfisher)
 import Kingfisher
 
-@available(iOS 14, *)
 private extension Image {
   func resolveWithKingFisher(from url: URL?, placeholder: Image?) -> some View {
     KFImage(url)
@@ -27,7 +25,6 @@ private extension Image {
 }
 #endif
 
-@available(iOS 14, *)
 extension Image {
   func resolve(from url: URL?, placeholder: Image?) -> some View {
     #if canImport(KingFisher)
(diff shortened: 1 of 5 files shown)


Sources/UI/ProfileImageView/ProfileImageView.swift
```diff
@@ -8,40 +8,46 @@
...
   
-  private func triggerImageTapClosure() { imageTapped?() }
-  private func tiggerBadgeTapClosure() { badgeTapped?() }
+  private func triggerImageTapClosure() { properties.imageTapped?() }
+  private func tiggerBadgeTapClosure() { properties.badgeTapped?() }
+}
+
+// MARK: - ViewModel
+private extension ProfileImageView {
+  class ProfileImageProperties: ObservableObject {
```
nagavcindriqim, line 36: Maybe we can rename this to `Properties`?

nagavcindriqim (approved):
Looks good 👌

Sources/UI/Extensions/Image+PovioKit.swift
```diff
@@ -8,13 +8,11 @@
 
 import SwiftUI
 
-@available(iOS 14, *)
```
borut-t, line 11: PovioKit currently supports iOS 13+. We'll probably bump soon to 14. Can we leave this intact?

borut-t (commented):
I thought we can also set the initials vs just an image. Have we lost this somehow?

Sources/UI/ProfileImageView/ProfileImageView.swift
```diff
@@ -178,15 +204,109 @@ public extension ProfileImageView {
...
+  var badging: BadgingMode {
+    get { properties.badging }
+    set { properties.badging = newValue }
+  }
+  
   func set(_ url: URL?) {
     properties.url = url
   }
+  
+  func addActionOnProfileTap(_ action: @escaping () -> Void) {
```
markom-povio, line 301: Maybe we should rename this just to be "onProfileTap" like on the above function?

Sources/UI/ProfileImageView/ProfileImageView.swift
```diff
@@ -178,15 +204,109 @@ public extension ProfileImageView {
...
+    properties.imageTapped = action
+  }
+  
+  func addActionOnBadgeTap(_ action: @escaping () -> Void) {
```
markom-povio, line 305: Also, this can be just "onBadgeTap"?

markom-povio (approved):
👍

Sources/UI/Extensions/Image+PovioKit.swift
ArlindDushi: @borut-t What do you mean by initials? Can you please give an example

borut-t: We can either set the image or initials inside the profile image view. Usually, we do this as a fallback when the user didn't provide picture.

ArlindDushi: We never had the implementation to set the initial properties. Was inspired by SwiftUI's `Image` as it has a simple init and then modifiers to add more specific requirements.
I can add nullable parameters to the init if we want to provide them from the get-go.
Personally wouldn't prefer it just because of style but I don't have strong feelings about this. Let me know what you think.

borut-t: I don't like that either. Maybe we've miss-understood. I just want to provide second option to image.

borut-t (approved):
👌

PR #270, opened 2023-04-14
Fix/Make url optional for SwiftUI

ArlindDushi:
Make url optional for SwiftUI so we don't have to force unwrap URL initialization


nagavcindriqim (approved):
🙌

borut-t (approved):

PR #275, opened 2023-04-20
Fix/Action button padding

ArlindDushi:
Removed left/right padding for title since we want it together with the image centered and not just center text.


borut-t (approved):
Great 👌

PR #280, opened 2023-05-08
Package re-structure

tonikocjan:
- Re-structured package:
   - PovioKitCore
     - Extensions/
   - PKNetworking
     - ANC/
   - PKPromiseKit
     - Promise/
   - PKSwiftUI
     -  Extensions/
     - ActionButton/
     - ProfileImageView/
   - PKUIKit
     - DynamicCollectionCell/
     - GradientView/
     - PaddingLabel
     - TextField
   - PKUtilities
     - _a lot of stuff_
- Support for MacOS


Package.swift
```diff
@@ -4,26 +4,55 @@ import PackageDescription
...
     .target(
       name: "PovioKitCore",
-      path: "Sources/Core"
+      path: "Sources/Core",
+      exclude: [
+        "../Networking",
+        "../PromiseKit",
+        "../Utilities",
+        "../UIKit",
+        "../SwiftUI",
```
borut-t, line 50: Do we need to explicitly exclude sources that are not included in the first place?

Sources/Utilities/InAppPurchase/InAppPurchaseService.swift
```diff
@@ -117,7 +116,7 @@ extension InAppPurchaseService {
   public func restorePurchases() async -> Result<Void, InAppPurchaseError> {
     do {
       try await AppStore.sync()
-      return .success()
+      return .success(())
```
borut-t, line 119: We have a `success()` Result extension that does just this.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
```diff
@@ -15,27 +15,21 @@ public extension Data {
...
-           .dataCorrupted(let context):
-        Logger.debug("Decoding (failure): \(type.self)", params: ["error": context.debugDescription])
-      @unknown default:
-        Logger.debug("Decoding (failure): \(type.self)", params: ["error": "Unknown decoding error."])
-      }
-      throw decodingError
-    } catch {
-      Logger.debug("Decoding (failure): \(type.self)", params: ["error": error.localizedDescription])
-      throw error
-    }
```
borut-t, line 39: Why removed?

Sources/Core/Extensions/Foundation/Encodable+PovioKit.swift
```diff
@@ -15,30 +15,17 @@ public extension Encodable {
...
-      switch encodingError {
-      case .invalidValue(_, let context):
-        Logger.debug("Encoding (failure): \(type(of: self))", params: ["error": context.debugDescription])
-      @unknown default:
-        Logger.debug("Encoding (failure): \(type(of: self))", params: ["error": "Unknown encoding error."])
-      }
-      throw encodingError
-    } catch {
-      Logger.debug("Encoding (failure): \(type(of: self))", params: ["error": error.self])
-      throw error
```
borut-t, line 40: Why removed?

Package.swift
```diff
@@ -4,26 +4,55 @@ import PackageDescription
...
+    .library(
+      name: "PovioKitSwiftUI", 
+      targets: ["PovioKitSwiftUI"]
+    ),
   ],
   dependencies: [
-    .package(url: "https://github.com/Alamofire/Alamofire", .upToNextMajor(from: "5.6.4"))
+    .package(
+      url: "https://github.com/Alamofire/Alamofire", 
+      .upToNextMajor(from: "5.6.4")
```
borut-t, line 38: Let's change to `5.0.0` instead.

Package.swift
```diff
@@ -4,26 +4,55 @@ import PackageDescription
...
+      targets: ["PovioKitPromise"]
+    ),
+    .library(
+      name: "PovioKitUIKit", 
+      targets: ["PovioKitUIKit"]
+    ),
+    .library(
+      name: "PovioKitSwiftUI", 
+      targets: ["PovioKitSwiftUI"]
+    ),
```
borut-t, line 33: I'm fine with naming. What I would change is the folder structure. I would have it like this:
- Sources
  - UI
    - UIKit
    - SwiftUI

borut-t (commented):
I like the new structure. The core is now small and lean 💪

Sources/Utilities/InAppPurchase/InAppPurchaseService.swift
tonikocjan: `PovioKitUtilities` does not depend on `PovioKitCore`. Wanted to make sub-packages have as few dependencies as possible.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
tonikocjan: 1. Logger is part of the `Utilities` package, do not want `Core` to depend on the subpackage.
2. Logging should be done on the client side.

Sources/Core/Extensions/Foundation/Encodable+PovioKit.swift
tonikocjan: Same as above.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
borut-t: Don't agree with both points.
1. True, but since the Logger is really the core of the whole package, I would leave it in the `Core` package. 
2. That should be part of the library. Client doesn't and shouldn't know about the implementation details.

Sources/Utilities/InAppPurchase/InAppPurchaseService.swift
borut-t: `Result` extension should be part of the core package 🤔

Sources/SwiftUI/ProfileImageView/ProfileImageView.swift
```diff
@@ -142,9 +155,11 @@ private extension ProfileImageView {
...
 
     
-    init?(style: BadgingMode,
-          size: CGSize,
-          badgeTapped: @escaping () -> ()) {
+    init?(
+      style: BadgingMode,
+      size: CGSize,
+      badgeTapped: @escaping () -> ()
+    ) {
```
borut-t, line 162: I don't want to negate other developers' code styles.

Sources/Utilities/InAppPurchase/InAppPurchaseService.swift
tonikocjan: `Result` is part of `Core,` but `InAppPurshaceService` is part of `Utilities`.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
tonikocjan: > That should be part of the library. Client doesn't and shouldn't know about the implementation details.

I disagree. A simple extension function such as this one should be completely pure, i.e. it should not modify the global state.

Sources/SwiftUI/ProfileImageView/ProfileImageView.swift
tonikocjan: I agree with your point and will revert. But I think we should strive to have a common style throughout the whole package.

Package.swift
tonikocjan: Apparently not. Removed.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
borut-t: I still tend to disagree. Extension is here for a reason, to "extend" the core functionality. By slashing pretty much all the code, the code left is meaningless, no gain for the developer. We have two options:
1) revert the change
2) remove the method completely

Sources/Core/Extensions/Foundation/Encodable+PovioKit.swift
borut-t: Same here. Either revert the change or remove the method completely.

Tests/Tests/Core/Utilities/Exif/ExifTests.swift
```diff
@@ -6,8 +6,10 @@
 //  Copyright © 2023 Povio Inc. All rights reserved.
 //
 
+#if os(iOS)
 import XCTest
 import PovioKitCore
+@testable import PovioKitUtilities
```
borut-t, line 12: Why using `@testable`? All APIs are public 🤔

borut-t (changes requested):
Great work. Although I have a few concerns:
- PR should re-structure the package, but is also doing other things
- PR is adding support for macos
- PR is changing existing code

We are doing three things in one PR which should be split into three separate PRs. I'm ok with first two points above to be done here, but let's scrap the third out into a separate PR.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
tonikocjan: The function itself is fine, it extends core functionality to use your lingo. But I would not force upon the user as to how they want to handle (and log the errors). 
If it comes to this I will just remove the method.

Tests/Tests/Core/Utilities/Exif/ExifTests.swift
tonikocjan: Removed where not necessary.

Sources/SwiftUI/ProfileImageView/ProfileImageView.swift
dejanskledar: I agree with @tonikocjan, can we  agree on one approach and enforce it throughout the team?

borut-t:
Great work @tonikocjan. We need some additional changes:
- rebase branch
- move `/Core/Utilities/MediaPlayer` to `Utilities` subpackage
- move `/Utilities/Logger` to `Core` subpackage (it's a must on every project, even if we don't need `Utilities` subpackage I should have access to the Logger from the Core subpackage)
- `Async` subpackage feels redundant to me. We have async code in other places as well, but we didn't move it here.

tonikocjan:
> Async subpackage feels redundant to me.

Apple also has a separate swift package for [Async data structures and algorithms](https://github.com/apple/swift-async-algorithms). I think separating makes sense as a user might only need some async algorithm that we provide. 

> We have async code in other places as well, but we didn't move it here.

The other async code is already nicely contained in the Networking and Promise sub-packages. Any other code that strongly utilizes Swift concurrency, should be moved inside the Async package.

borut-t:
> > Async subpackage feels redundant to me.
> 
> Apple also has a separate swift package for [Async data structures and algorithms](https://github.com/apple/swift-async-algorithms). I think separating makes sense as a user might only need some async algorithm that we provide.
> 
> > We have async code in other places as well, but we didn't move it here.
> 
> The other async code is already nicely contained in the Networking and Promise sub-packages. Any other code that strongly utilizes Swift concurrency, should be moved inside the Async package.

Great. Let's proceed 👌

borut-t (approved):
🙌

PR #288, opened 2023-08-10
Media Player & Audio Player

dejanskledar:
- Updated Media Player and expanded the Interface
- Added Audio Player logic and feature


borut-t (approved):
Great work. I've done some additional changes in the follow-up PR.

tonikocjan (approved):

PR #291, opened 2023-10-19
Make Alamofire URL Session public for easier customization

dejanskledar:
With this, we enable runtime dynamic customization of the default Session, after creation which was currently not possible.

### DIFF
--- a/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
+++ b/Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
@@ -22,7 +22,7 @@ public typealias ProgressHandler = Alamofire.Request.ProgressHandler
 public typealias ErrorHandler = (Swift.Error, Data) throws -> Swift.Error
 
 open class AlamofireNetworkClient {
-  private let session: Alamofire.Session
+  public let session: Alamofire.Session
   private let eventMonitors: [RequestMonitor]
   private let defaultErrorHandler: ErrorHandler?
   


borut-t (approved):
This sounds reasonable to me.

Sources/Networking/AlamofireNetworkClient/AlamofireNetworkClient.swift
```diff
@@ -22,7 +22,7 @@ public typealias ProgressHandler = Alamofire.Request.ProgressHandler
 public typealias ErrorHandler = (Swift.Error, Data) throws -> Swift.Error
 
 open class AlamofireNetworkClient {
-  private let session: Alamofire.Session
+  public let session: Alamofire.Session
```
tonikocjan, line 25: You could create a new client with a different session instead. Would that work for your use-case?

dejanlatecheckout: That was the initial idea, but for our use case its not possible without changing the core logic of our Network Layer / Package. :/

borut-t: Thinking.... 🤔
We could eventually inject required state to the project package instead, and go without this as well. What do you think @dejanlatecheckout?

dejanlatecheckout: This would work, let me try it out.

dejanskledar:
Found a workaround, closing.

PR #294, opened 2023-10-27
Feature/Implement tappable SwiftUI Text that redirects to link

yllfejziu:


Tests/Tests/Core/Extensions/SwiftUI/TextTests.swift
markom-povio, line 1: Do we need this test in the `Core` package?
Maybe we can add this component to the Storybook demo project and add UI tests there instead. What do you think?

yllfejziu: I agree, let's omit it from `Core` for now.

markom-povio (approved):
👍

Sources/Core/Extensions/SwiftUI/Text+PovioKit.swift
```diff
@@ -0,0 +1,21 @@
...
+//  PovioKit
+//
+//  Created by Yll Fejziu on 27/10/2023.
+//  Copyright © 2023 Povio Inc. All rights reserved.
+//
+
+import SwiftUI
+
+public extension Text {
+  /// Creates an instance of hypelinked Text using markdown formatting syntac
```
borut-t, line 12: Typo "hypelinked".

borut-t (approved):
👌

PR #295, opened 2023-11-01
Feature/OSLog

borut-t:
Replaced standard output mechanism a.k.a `print` with the modern performant os-level logging system, for iOS 14 and above. Here are the benefits:

- low performance overhead
- logs are persisted on the device for certain period
- you can access logs using the Console app

Additionally output was formatted a bit.
```swift
Logger.error("This is a message with params",
             params: ["error": "Input data is invalid!", "userId": 12345, "code": 500])
```
This code will output the following in the console:

before...
```swift
ERROR: AppDelegate.application(_:didFinishLaunchingWithOptions:)[31]: This is a message with params, params: [\"userId\": 12345, \"code\": 500, \"error\": \"Input data is invalid!\"]
```

after...
```swift
ERROR: This is a message with params
 ⮑ source: AppDelegate.application(_:didFinishLaunchingWithOptions:):31
 ⮑ code: 500
 ⮑ error: Input data is invalid!
 ⮑ userId: 12345
```

### DIFF
--- a/Sources/Core/Utilities/Logger/Logger.swift
+++ b/Sources/Core/Utilities/Logger/Logger.swift
@@ -64,11 +64,6 @@ public extension Logger {
   static func error(_ message: String, params: Parameters? = nil, file: String = #file, function: String = #function, line: Int = #line) {
     flush(.error, message: message, params: params, file: file, function: function, line: line)
   }
-  
-  /// Log crash
-  static func crash(params: Parameters, file: String = #file, function: String = #function, line: Int = #line) {
-    flush(.error, message: "app_crash", params: params, file: file, function: function, line: line)
-  }
 }
 
 // MARK: - Private Methods
--- a/Sources/Core/Utilities/Logger/Logger.swift
+++ b/Sources/Core/Utilities/Logger/Logger.swift
@@ -75,7 +75,7 @@ private extension Logger {
     let nl = "\n ⮑ "
     var messagePrint = "\(level.label): \(message)"
     if line >= 0 {
-      messagePrint += "\(nl)source: \(fileName).\(function)[\(line)]"
+      messagePrint += "\(nl)source: \(fileName).\(function):\(line)"
     }
     if let params, !params.isEmpty {
       let groupedParams = params
--- a/Sources/Core/Utilities/Logger/Logger.swift
+++ b/Sources/Core/Utilities/Logger/Logger.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2023 Povio Inc. All rights reserved.
 //
 
-import Foundation
+import OSLog
 
 public final class Logger {
   public typealias Parameters = [String: Any]
@@ -26,7 +26,7 @@ public extension Logger {
     case debug
     case all
     
-    var string: String {
+    var label: String {
       switch self {
       case .info:
         return "INFO"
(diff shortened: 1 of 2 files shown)


Sources/Networking/AlamofireNetworkClient/AlamofireConsoleLogger.swift
```diff
@@ -14,7 +14,7 @@ public final class AlamofireConsoleLogger: EventMonitor {
   public let queue = DispatchQueue(label: "com.alamofire.console.networklogger")
   private let logger: (String) -> Void
   
-  public init(logger: @escaping (String) -> Void = { Logger.debug($0) }) {
+  public init(logger: @escaping (String) -> Void = { Logger.debug($0, line: -1) }) {
```
markom-povio, line 17: Is this intentional or is it a leftover?

markom-povio (approved):
Nice 👌

Sources/Networking/AlamofireNetworkClient/AlamofireConsoleLogger.swift
borut-t: That is intentional. If you take a look at [this line](https://github.com/poviolabs/PovioKit/pull/295/files#diff-e17de7f55086084fc2cab7b1efeeea7e81eea4033353c1af71c92b0e4e53ddc9R77), you'll notice we don't output source metadata when line is below 0. In this case we try to omit source output since it's always pointing to the `AlamofireConsoleLogger`, thus irrelevant. Hope it makes sense. If you have a better suggestion, lmk.

Sources/Core/Utilities/Logger/Logger.swift
```diff
@@ -64,25 +64,47 @@ public extension Logger {
...
-    debugPrint(messagePrint)
-    debugPrint()
+    if #available(iOS 14.0, *) {
+      let category = "\(fileName) - \(function) - line \(line)"
+      let logger = os.Logger(subsystem: Bundle.main.bundleIdentifier ?? "povioKit.logger", category: category)
+      switch level {
+      case .none:
+        break
+      case .error:
+        logger.error("\(messagePrint)")
```
EgzonArifi, line 95: we can pass directly pass `messagePrint` without string interpolation

Sources/Core/Utilities/Logger/Logger.swift
```diff
@@ -64,25 +64,47 @@ public extension Logger {
...
 // MARK: - Private Methods
 private extension Logger {
   static func flush(_ level: LogLevel, message: String, params: Parameters? = nil, file: String, function: String, line: Int) {
     guard shared.logLevel.rawValue >= level.rawValue else { return }
     
-    let fileName = URL(fileURLWithPath: file).lastPathComponent.components(separatedBy: ".").first ?? ""
-    var messagePrint = "\(level.string): \(fileName).\(function)[\(line)]: \(message)"
-    if let params = params, !params.isEmpty {
-      messagePrint += ", params: \(params)"
+    let fileName = URL(string: file)?.lastPathComponent ?? ""
```
EgzonArifi, line 74: I think we still should use `URL(fileURLWithPath: file)` , to ensure the path is correctly interpreted as a file path

Sources/Networking/AlamofireNetworkClient/AlamofireConsoleLogger.swift
markom-povio: I see, it makes sense...

Sources/Core/Utilities/Logger/Logger.swift
borut-t: Not really - https://developer.apple.com/documentation/os/oslogmessage

Sources/Core/Utilities/Logger/Logger.swift
borut-t: Good point 👌

PR #305, opened 2024-02-06
Broadcast/Minor refactoring and optimization

tonikocjan:
- Fixed compilation warning (error in Swift 6)
- Performed minor code improvements
- Remove observer optimization

### DIFF
--- a/Sources/Core/Utilities/Broadcast/Broadcast.swift
+++ b/Sources/Core/Utilities/Broadcast/Broadcast.swift
@@ -9,13 +9,13 @@
 import Foundation
 
 public final class Broadcast<T> {
-  private(set) var observers = [Weak<AnyObject>]()
+  private(set) var observers = [Weak]()
 
   public init() {}
   
   public func add(observer: T) {
     prune()
-    observers.append(Weak<AnyObject>(observer as AnyObject))
+    observers.append(Weak(observer as AnyObject))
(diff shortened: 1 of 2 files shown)


Sources/Core/Utilities/Broadcast/Broadcast.swift
```diff
@@ -3,55 +3,68 @@
 //  PovioKit
 //
 //  Created by Domagoj Kulundzic on 26/04/2019.
-//  Copyright © 2023 Povio Inc. All rights reserved.
+//  Copyright © 2024 Povio Inc. All rights reserved.
```
borut-t, line 6: @tonikocjan can you prepare a PR updating all headers year?

borut-t (approved):
👌

Sources/Core/Utilities/Broadcast/Broadcast.swift
```diff
@@ -3,55 +3,68 @@
...
   }
   
   public func clear() {
-    delegates.removeAll()
+    observers.removeAll()
+  }
+}
+
+extension Broadcast {
+  class Weak {
```
tonikocjan, line 57: Actually, there is no need for `Weak` to be generic at all, hence I removed the generic parameter.

borut-t (approved):

PR #309, opened 2024-03-08
AsyncThrottleSequence/Fixed error propagation

tonikocjan:
Fixed the issue where error thrown by the throttled operation was not propagated to the caller.


markom-povio (approved):

borut-t (approved):
Much cleaner 👌

PR #310, opened 2024-03-18
Reimplemented `count(where: )`

tonikocjan:


Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
```diff
@@ -17,8 +17,14 @@ public extension Collection where Indices.Iterator.Element == Index {
...
-  func count(where clause: (Element) -> Bool) -> Int {
-    lazy.filter(clause).count
+  func count(where predicate: (Element) throws -> Bool) rethrows -> Int {
+    var count = 0
+    for el in self {
+      if try predicate(el) {
+        count += 1
+      }
+    }
+    return count
```
borut-t, line 27: Where do you see the benefit?

seckmaster: Using `lazy` in such cases is not the appropriate mechanism.
The implementation in this PR is the same as the [original implementation which was accepted into the Stdlib](https://github.com/apple/swift/pull/16099/files#diff-b7c882e9c2a82394d1b3a544fbb86539a5aaf137dea62ba27aaf5200bfc249d4R550) though later removed.

However, I am considering implementing it by using `reduce` instead (reference: https://forums.swift.org/t/refresh-review-se-0220-count-where/66235/4).

tonikocjan: Using `lazy` in such cases is not the appropriate mechanism.
The implementation in this PR is the same as the [original implementation which was accepted into the Stdlib](https://github.com/apple/swift/pull/16099/files#diff-b7c882e9c2a82394d1b3a544fbb86539a5aaf137dea62ba27aaf5200bfc249d4R550) though later removed.

However, I am considering implementing it by using `reduce` instead (reference: https://forums.swift.org/t/refresh-review-se-0220-count-where/66235/4).

yllfejziu (approved):
👌

Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
```diff
@@ -16,15 +16,18 @@ public extension Collection where Indices.Iterator.Element == Index {
...
-    for el in self {
-      if try predicate(el) {
-        count += 1
+  /// Conditional element count - https://forums.swift.org/t/refresh-review-se-0220-count-where/66235/4
+  @inlinable
+  func count(
+    where predicate: (Element) throws -> Bool
+  ) rethrows -> Int {
+    try reduce(0) { n, element in
+      if try predicate(element) {
```
borut-t, line 25: Great. Reduce is indeed much faster operation 👌

borut-t (approved):

PR #313, opened 2024-04-03
`mutateEach(:)`

tonikocjan:
Implemented a mutable version of `forEach`.

Instead of writing smth like

```swift
for idx in 0..<c.count {
  c[idx].mutate()
}
```
, you can use `mutateEach`:

```swift
c.mutateEach { $0.mutate() }
```


borut-t (approved):
👌

yllfejziu (approved):

ArminSpahic (approved):

PR #319, opened 2024-05-13
Feature/Pinch to zoom

yllfejziu:


Sources/UI/SwiftUI/View Modifiers/PinchToZoomModifier.swift
```diff
@@ -0,0 +1,164 @@
...
+    didSet {
+      delegate?.pinchZoomView(self, didChangeOffset: offset)
+    }
+  }
+
+  private(set) var isPinching: Bool = false {
+    didSet {
+      delegate?.pinchZoomView(self, didChangePinching: isPinching)
+    }
+  }
```
tonikocjan, line 73: I would avoid the delegate pattern and modify the view's state directly instead.

borut-t (approved):
Agree with @tonikocjan.

Sources/UI/SwiftUI/View Modifiers/PinchToZoomModifier.swift
yllfejziu: We’re using UIViewRepresentable here, and if I understand correctly, your suggestion is to bind the UIKit view state directly to SwiftUI state. 
One way to achieve this would be using Combine, as it allows us to directly bind state changes without a delegate pattern. Is this what you had in mind, or did you have another approach in mind for avoiding the delegate pattern?

tonikocjan: No need to use Combine. One solution is to move the state into a view model and inject a reference to the view model into `PinchZoomView`.

yllfejziu:
Demo of the feature in povio app:

https://github.com/poviolabs/PovioKit/assets/38266916/f45e6995-9eb7-4f9c-b2ea-2e8d6657fe70

nagavcindriqim (approved):
Works great 🙆‍♂️

tonikocjan (approved):
👍

Sources/UI/SwiftUI/View Modifiers/PinchToZoomModifier.swift
```diff
@@ -0,0 +1,109 @@
...
+      .scaleEffect(viewModel.scale, anchor: viewModel.anchor)
+      .offset(viewModel.offset)
+      .animation(viewModel.isPinching ? .none : .spring(), value: viewModel.scale)
+      .overlay(
+        PinchZoom(viewModel: viewModel)
+      )
+  }
+}
+
+@available(iOS 14.0, *)
```
borut-t, line 27: We don't really want to support SwiftUI on iOS 14. Only 15+. Please update.

borut-t (changes requested):

Sources/UI/SwiftUI/View Modifiers/PinchToZoomModifier.swift
yllfejziu: Updated

borut-t (approved):
👌

PR #327, opened 2024-06-14
Utilities/ImageTools

borut-t:
A set of image tools bundled in a handy `UIImage` extension. Some methods depends on the Kingfisher library, therefore it needs to be installed in the project in order to leverage them.


Sources/Utilities/Image Saver/ImageTools.swift
```diff
@@ -0,0 +1,301 @@
...
+    switch format {
+    case .jpeg(let compressionRatio):
+      compressedImage = image.jpegData(compressionQuality: compressionRatio)
+    case .png:
+      compressedImage = image.pngData()
+    }
+    
+    guard let compressedImage else { throw ImageError.compression }
+    
+    print("Image compressed to:", Double(compressedImage.count) / 1024.0 / 1024.0, "MB")
```
dejanskledar, line 168: Use Logger here?

Sources/Utilities/Image Saver/ImageTools.swift
```diff
@@ -0,0 +1,301 @@
...
+  /// ```swift
+  /// do {
+  ///   let image = UIImage(named: "exampleImage")!
+  ///   let compressedData = try compress(image, format: .jpeg(compressionRatio: 0.8))
+  /// } catch {
+  ///   print("Failed to compress image: \(error)")
+  /// }
+  /// ```
+  /// - Throws: `ImageError.compression` if the compression operation fails.
+  func compress(_ image: UIImage, format: ImageFormat) throws -> Data {
```
dejanskledar, line 157: Lets use swift syntax?:

```swift
func compress(image: UIImage, with format: ImageFormat)
```

Sources/Utilities/Image Saver/ImageTools.swift
```diff
@@ -0,0 +1,301 @@
...
+  ///   let percentage: CGFloat = 50.0
+  ///   let resizedImage = try await downsize(image, by: percentage)
+  ///   imageView.image = resizedImage
+  /// } catch {
+  ///   print("Failed to downsize image: \(error)")
+  /// }
+  /// ```
+  /// - Note: This function should be called from an asynchronous context using `await`.
+  /// - Throws: `ImageError.invalidPercentage` if the percentage is not within the valid range.
+  func downsize(_ image: UIImage, by percentage: CGFloat) async throws -> UIImage {
```
dejanskledar, line 116: Lets use swift syntax?:
```swift
  func downsize(image: UIImage, by percentage: CGFloat) async throws -> UIImage {
```

Sources/Utilities/Image Saver/ImageTools.swift
```diff
@@ -0,0 +1,301 @@
...
+  /// - Returns: An optional UIImage that is the result of the downsizing operation.
+  /// - Example:
+  /// ```swift
+  /// let image = UIImage(named: "exampleImage")!
+  /// let targetSize = CGSize(width: 100, height: 100)
+  /// let resizedImage = await downsize(image, to: targetSize)
+  /// imageView.image = resizedImage
+  /// ```
+  /// - Note: This function should be called from an asynchronous context using `await`.
+  func downsize(_ image: UIImage, to targetSize: CGSize) async -> UIImage {
```
dejanskledar, line 77: ditto

Sources/Utilities/Image Saver/ImageTools.swift
```diff
@@ -0,0 +1,301 @@
...
+  ///   let image = UIImage(named: "exampleImage")!
+  ///   try await saveImageToPhotoLibrary(image)
+  ///   print("Image saved successfully.")
+  /// } catch {
+  ///   print("Failed to save image: \(error)")
+  /// }
+  /// ```
+  /// - Note: This function should be called from an asynchronous context using `await`.
+  /// - Throws: An error if the save operation fails.
+  func saveImageToPhotoLibrary(_ image: UIImage) async throws {
```
dejanskledar, line 52: ditto

dejanskledar (commented):
Godo job!

Another addition, would be to compress/reduce size to a max size, eg. if server expects an image that is max 1MB, a nice method would be 

```swift
func compress(image: UIImage, toMaxKbSize size: CGFloat) throws -> Data 
```

Also, lets use Swift syntax with labeled parameters? Its such a great concept Swift offers, its really a shame we omit it, feels like a step back 😅

Sources/Utilities/Image Saver/ImageTools.swift
borut-t: I agree this should be unified.

The call sites...
```swift
now
let data = compress(myImage, format: .png)

after
let data = compres(image: myImage, with: .png)
```

Your proposal changes first param to explicit, where the second is implicit/unclear.
I my case, the first is implicit, the second is explicit.
If you explicitly define one param, we should do the same for others.

With that in mind, you suggestion should be modified to
```swift
func compress(image: UIImage, withFormat: ImageFormat)
```

I would change it to
```swift
func compress(image: UIImage, format: ImageFormat)
// or
func compress(image: UIImage, withFormat format: ImageFormat)
```

The above solutions call site is as follows:
```swift
compress(image: myImage, format: .png)
compress(image: myImage, withFormat: .png)
```

dejanskledar: Perfect, completely agree, makes more sense 👏

Sources/Utilities/ImageTools/ImageTools.swift
```diff
@@ -0,0 +1,358 @@
...
+        height: floor(originalSize.height * scaleFactor)
+      )
+      
+      let renderer = UIGraphicsImageRenderer(size: newSize)
+      let newImage = renderer.image { _ in
+        image.draw(in: CGRect(origin: .zero, size: newSize))
+      }
+      
+      return newImage
+    }.value
```
tonikocjan, line 104: I do not like making assumptions about how the user will use the code. Why wrap the synchronous code into an async block, especially with a detached task? 
The concurrency overhead might be higher than the actual image rendering.

Sources/Utilities/ImageTools/ImageTools.swift
```diff
@@ -0,0 +1,358 @@
...
+    try await withCheckedThrowingContinuation { continuation in
+      let continuationWrapper = ContinuationWrapper(continuation: continuation)
+      UIImageWriteToSavedPhotosAlbum(
+        image,
+        self,
+        #selector(saveCompleted),
+        Unmanaged.passRetained(continuationWrapper).toOpaque()
+      )
+    }
+  }
```
tonikocjan, line 64: Same here, I would leave it as synchronous.

Sources/Utilities/ImageTools/ImageTools.swift
```diff
@@ -0,0 +1,358 @@
...
+/// A utility class for performing various image-related operations.
+public final class ImageTools: NSObject {
+#if canImport(Kingfisher)
+  private var prefetcher: ImagePrefetcher?
+#endif
+  
+  override public init() {
+    super.init()
+  }
+}
```
tonikocjan, line 21: Do we need this or can we just extend `UIImage`? 🤔

Sources/Utilities/ImageTools/ImageTools.swift
dejanskledar: Good thought... not that you mention it, I never to async in image resizing ... 🤔

borut-t: Image processing tasks are rather expensive and will block the main thread if not used correctly. Which we want to avoid at all cost.

I agree with you to a point, that the call site should handle that. That's the go-to solution for many apis.

In most cases, we are processing larger images that do take portion of the time to process and blocks the main thread. Using this api, dev doesn't need to think how to properly use it not to block the UI.

Similar to how Kingfisher works. You don't actually care how it does it, you just need the result without sacrificing user experience.

We can go the other direction and have all this apis sync, leave that decision to the caller and don't care about the result. Which will in most cases result to a bad ux. We all know we tend to be lazy, forget about the performance when we are mostly testing on the latest, most performant hardware. And we only realise the issue when the customers reports laggy experience with older devices. 
On the other hand by "enforcing" user async context, it will never lead to such situation.

tonikocjan: Are you sure `Task.detached` is the solution? As far as I can tell, the documentation does not mention on which thread the newly created task would execute. For all you know, the code can be executed on the main thread.

borut-t: You are correct. You cannot define the thread on which Task will run. You can instead define the priority only. And the Task itself will figure out what's the best thread to run on based on the available system resources. We never actually care about the thread, but the queues (concurrency runs on top of GCD).

One thing I need to change it to remove the detached task. It's not needed and in many cases less safe than regular Task.

ArlindDushi (approved):
Great work 👌

Sources/Utilities/ImageTools/ImageTools.swift
borut-t: Cannot do that. We need to have an async code just to support old selector based api.

Sources/Utilities/ImageTools/ImageTools.swift
borut-t: We need it, it's used in the `prefetch` method.

tonikocjan: - saveToPhotoLibrary
- downsize
- compress
- ...

could just be UIImage extensions, or am I missing something?

borut-t: Good point. It makes sense to have UIImage extension instead. I've separated into two. `UIImage+PovioKit` where all UIKit operations are happening and `UIImage+Kingfisher` for library specific operations.

PR #328, opened 2024-08-07
Chore/Privacy manifest relocation

borut-t:


dejanskledar (approved):
🚀

EgzonArifi (commented):
Question: why not keeping it under `Resources`? 🤔

borut-t:
> Question: why not keeping it under `Resources`? 🤔

Good question @EgzonArifi. I think this should be part of the Source and not the Resource due to the nature of this file. And the second thing is the nesting, I don't want that tbh.

EgzonArifi (approved):

PR #335, opened 2024-10-02
Extensions

borut-t:
Added new small but useful extensions.


Sources/Core/Extensions/Foundation/Array+PovioKit.swift
```diff
@@ -0,0 +1,55 @@
...
+  /// ## Example
+  /// ```
+  /// let array = [1, 2, 3, 4, 5, 6, 7]
+  /// let chunks = array.chunked(into: 3)
+  /// // chunks is [[1, 2, 3], [4, 5, 6], [7]]
+  /// ```
+  func chunked(into size: Int) -> [[Element]] {
+    stride(from: 0, to: count, by: size).map {
+      Array(self[$0 ..< Swift.min($0 + size, count)])
+    }
```
tonikocjan, line 28: It would be much faster to return a list of subsequences.

tonikocjan: We can also make this method more general by extending `Collection`.

Sources/Core/Extensions/Foundation/Array+PovioKit.swift
```diff
@@ -0,0 +1,55 @@
...
+  /// let failure = array.update(10, at: 10)
+  /// // array remains [1, 2, 10, 4]
+  /// // failure is false
+  /// ```
+  @discardableResult
+  mutating func update(_ item: Element, at index: Int) -> Bool {
+    guard indices.contains(index) else { return false }
+    self[index] = item
+    return true
+  }
```
tonikocjan, line 54: 1. I would maybe call it `safeUpdate`
2. I would extend `RandomAccessCollection` instead.

Sources/Core/Extensions/Foundation/String+PovioKit.swift
```diff
@@ -31,6 +31,16 @@ public extension String {
     data(using: .utf8)?.base64EncodedString(options: Data.Base64EncodingOptions(rawValue: 0))
   }
   
+  /// Returns an array of strings, where each string represents a line from the original string, split at newline characters.
+  var lines: [String] {
+    split(whereSeparator: \.isNewline).map(String.init)
```
tonikocjan, line 36: Returning a list of substrings would be much faster.

Sources/Core/Extensions/Foundation/String+PovioKit.swift
```diff
@@ -31,6 +31,16 @@ public extension String {
...
+  }
+  
+  /// Returns a boolean value indicating whether the string contains any emoji characters.
+  var containsEmoji: Bool {
+    unicodeScalars.contains { $0.properties.isEmojiPresentation }
+  }
```
tonikocjan, line 42: Do we need this? Look very specific.

Sources/Core/Extensions/Foundation/URL+PovioKit.swift
```diff
@@ -32,9 +34,24 @@ public extension URL {
...
+  /// - Returns: An optional dictionary where the keys are `AnyHashable` representing the query parameter names,
+  ///            and the values are `Any` representing the corresponding query parameter values.
+  ///            Returns nil if the URL is invalid or has no query parameters.
+  var queryParameters: [AnyHashable: Any]? {
+    guard let components = URLComponents(url: self, resolvingAgainstBaseURL: true),
+          let queryItems = components.queryItems else { return nil }
+    var params: [AnyHashable: Any] = [:]
+    queryItems.forEach { queryItem in
+      queryItem.value.map { params[queryItem.name] = $0 }
+    }
```
tonikocjan, line 49: Use regular for loop or `reduce`.

tonikocjan: Reserving capacity in advance is also not a bad idea.

Sources/Core/Extensions/Foundation/DateFormatter+PovioKit.swift
```diff
@@ -0,0 +1,91 @@
...
+    return dateFormatter
+  }()
+  
+  /// Format a time in a 24-hour format without AM/PM designation.
+  ///
+  /// `9:41`
+  static let time24Hour: DateFormatter = {
+    let dateFormatter = DateFormatter()
+    dateFormatter.locale = .autoupdatingCurrent
+    dateFormatter.dateFormat = "h:mm"
```
ArlindDushi, line 28: I'm guessing we'd want `HH:mm` format here otherwise if it's 15:03 it translates to 3:03 and we don't know if it's AM or PM

Sources/Core/Extensions/Foundation/URL+PovioKit.swift
```diff
@@ -32,9 +34,24 @@ public extension URL {
...
+    return params
+  }
 }
 
-extension URL: ExpressibleByStringLiteral {
+extension URL: @retroactive ExpressibleByStringLiteral {
```
ArlindDushi, line 54: Can we use `Swift.ExpressibleByStringLiteral` so this code can still build on older swift versions

ArlindDushi (commented):
Nice work 👌 . Please check a few nitpicks

Sources/Core/Extensions/Foundation/Array+PovioKit.swift
borut-t: RandomAccessCollection is indeed the best choice 👌

Sources/Core/Extensions/Foundation/Array+PovioKit.swift
borut-t: I will leave just `update` because safe doesn't bring nothing. We are always returning the success status.

Sources/Core/Extensions/Foundation/String+PovioKit.swift
borut-t: Not often, but usable to determine if string contains an emoji.

Sources/Core/Extensions/Foundation/URL+PovioKit.swift
borut-t: 👍

Sources/Core/Extensions/Foundation/DateFormatter+PovioKit.swift
borut-t: Thanks for the notice. Your solution is indeed better. But I would only have one Hour component to not show leading zeroes.

Sources/Core/Extensions/Foundation/URL+PovioKit.swift
borut-t: Tbh, I don't like it. And regarding to the [documentation](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0364-retroactive-conformance-warning.md#source-compatibility), we can do it differently.

Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
```diff
@@ -56,3 +158,59 @@ public extension MutableCollection {
...
+  /// let success = array.update(10, at: 2)
+  /// // array is now [1, 2, 10, 4]
+  /// // success is true
+  ///
+  /// let failure = array.update(10, at: 10)
+  /// // array remains [1, 2, 10, 4]
+  /// // failure is false
+  /// ```
+  @discardableResult
+  mutating func update(_ item: Element, at index: Index) -> Bool {
```
yllfejziu, line 182: Would it make sense to return the updated item as Element instead of Bool? This way, nil could indicate an invalid index, potentially adding flexibility for chaining operations.

Sources/Core/Extensions/Foundation/Data+PovioKit.swift
```diff
@@ -9,25 +9,63 @@
...
+  /// Useful for printing out device tokens, hashing results, or any scenario where raw data needs to be displayed as a hexadecimal string.
+  ///
+  /// - Returns: A string containing the hexadecimal representation of the data.
+  ///
+  /// ## Example
+  /// ```swift
+  /// let data = Data([0x01, 0xAB, 0xFF])
+  /// Logger.debug(data.hexadecimalString) // "01abff"
+  /// ```
   var hexadecialString: String {
```
yllfejziu, line 25: Let's fix typo: hexadecial

yllfejziu (commented):
Great work 👌
Couple comments below 👇

Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
borut-t: You are passing in an item to update. I understand your notion here, but I think it doesn't make much sense tbh.

ArlindDushi (approved):

tonikocjan (approved):

markom-povio (approved):

yllfejziu (approved):

Sources/Core/Extensions/Foundation/Collection+PovioKit.swift
```diff
@@ -10,17 +10,42 @@ import Foundation
...
+  /// - Returns: The element at the specified index if within bounds, otherwise `nil`.
+  ///
+  /// ## Example:
+  /// ```swift
+  /// let array = [1, 2, 3, 4]
+  /// let element = array[safe: 2] // returns 3
+  /// let outOfBounds = array[safe: 10] // returns nil
+  /// ```
+  subscript(safe index: Index) -> Element? {
     if startIndex <= index && index < endIndex { self[index] } else { nil }
```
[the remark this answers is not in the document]
borut-t: Performance wise, this is equal to the current solution. But I like the readability aspect 👌

PR #337, opened 2024-12-24
Fix/RemoteImage

borut-t:
Robust RemoteImage for loading remote images either with Kingfisher (if imported) or with system AsyncImage.


yllfejziu (approved):
👌

dejanskledar (approved):
Nice catch

PR #343, opened 2025-02-25
Fix/UserDefault migration

borut-t:
Fixed an issue where migration from default UserDefaults value storage to codable storage failed.


gapi (approved):
Migration issue fixed 👌

PR #347, opened 2025-03-10
Fix/Markdown

JanFicko:
This PR fixes markdown interpretation to include elements that were not displayed correctly previously:
- headers
- lists
- code blocks
- blockquotes
- tables

### DIFF
--- a/Sources/Core/Extensions/Foundation/String+PovioKit.swift
+++ b/Sources/Core/Extensions/Foundation/String+PovioKit.swift
@@ -83,7 +83,8 @@ public extension String {
   @available(iOS 15, *)
   func toMarkdown() -> AttributedString {
     do {
-      return try AttributedString(markdown: self)
+      let options = AttributedString.MarkdownParsingOptions(interpretedSyntax: .full)
+      return try AttributedString(markdown: self, options: options)
     } catch {
       return AttributedString(self)
     }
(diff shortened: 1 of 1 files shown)


Sources/Core/Extensions/Foundation/String+PovioKit.swift
```diff
@@ -83,7 +83,8 @@ public extension String {
   @available(iOS 15, *)
   func toMarkdown() -> AttributedString {
     do {
-      return try AttributedString(markdown: self)
+      let options = AttributedString.MarkdownParsingOptions(interpretedSyntax: .full)
+      return try AttributedString(markdown: self, options: options)
```
borut-t, line 87: Have you tested it out on how rendering improves with this change?

borut-t (approved):
🙌

Sources/Core/Extensions/Foundation/String+PovioKit.swift
JanFicko: Yes, I've tested it on a project where we use markdown elements. It now works with new lines, bulleted/numbered list, code block, ..., which was previously ignored.

Sources/Core/Extensions/Foundation/String+PovioKit.swift
```diff
@@ -83,7 +83,8 @@ public extension String {
   @available(iOS 15, *)
   func toMarkdown() -> AttributedString {
     do {
-      return try AttributedString(markdown: self)
+      let options = AttributedString.MarkdownParsingOptions(interpretedSyntax: .full)
```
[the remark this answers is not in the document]
borut-t: Good point. We can have them as param defaulting to this options ...

ArlindDushi (approved):
👌

dejanskledar (approved):
🙌

PR #352, opened 2025-04-30
Fix/Changing the safari schema to > ios 17.5 for openUrl

darshanbhatta:
I noticed an issue on IOS 17.0 where the `openUrl` function when `inSafari=true` wouldn't open the URL due to the safari schema not working on that version. It turns out it only is supported on ios 17.5 and above.

Updated the util function to have the correct versioning logic in place


borut-t (approved):
Great find @darshanbhatta 🙌

PR #356, opened 2025-10-03
Chore/Kingfisher as dependecy

borut-t:
Made Kingfisher part of the PovioKit.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -75,7 +75,6 @@ let package = Package(
     .target(
       name: "PovioKitUIKit",
       dependencies: [
-        "Kingfisher",
         "PovioKitCore",
         "PovioKitUtilities",
       ],
@@ -85,7 +84,6 @@ let package = Package(
     .target(
       name: "PovioKitSwiftUI",
       dependencies: [
-        "Kingfisher",
         "PovioKitCore",
       ],
       path: "Sources/UI/SwiftUI",
--- a/Package.swift
+++ b/Package.swift
@@ -60,6 +60,7 @@ let package = Package(
       name: "PovioKitNetworking",
       dependencies: [
         "Alamofire",
+        "PovioKitCore",
         "PovioKitPromise",
       ],
       path: "Sources/Networking",
--- a/Package.swift
+++ b/Package.swift
@@ -41,11 +41,18 @@ let package = Package(
     .package(
       url: "https://github.com/Alamofire/Alamofire", 
       .upToNextMajor(from: "5.0.0")
+    ),
+    .package(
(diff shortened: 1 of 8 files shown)


Package.swift
```diff
@@ -67,6 +75,7 @@ let package = Package(
     .target(
       name: "PovioKitUIKit",
       dependencies: [
+        "Kingfisher",
```
tonikocjan: Not sure why this is needed if this subpackage already depends on `PovioKitCore`?

Package.swift
```diff
@@ -76,6 +85,7 @@ let package = Package(
     .target(
       name: "PovioKitSwiftUI",
       dependencies: [
+        "Kingfisher",
```
tonikocjan: Same here?

Sources/UI/SwiftUI/Views/RemoteImage/RemoteImage.swift
```diff
@@ -32,38 +30,45 @@ public struct RemoteImage<Placeholder: View>: View {
...
   private var placeholder: Placeholder?
+  private var onSuccess: ((RetrieveImageResult) -> Void)?
+  private var onError: ((KingfisherError) -> Void)?
   
-  public init(url: URL?, animated: Bool = false) where Placeholder == EmptyView {
+  public init(
+    url: URL?,
+    animated: Bool = false,
+    onSuccess: ((RetrieveImageResult) -> Void)? = nil,
+    onError: ((KingfisherError) -> Void)? = nil
```
tonikocjan: Maybe we can apply the same API design as `KFImage`?
So that the usage would be:

```swift
RemoteImage()
  .onSuccess {}
  .onFailure {}
```

What do you think?

Package.swift
borut-t: [Redundant](https://github.com/povio/PovioKit/pull/356/commits/fd642483f4cb65cdbb467f8be60dbd8707d3a067) yes.

Package.swift
borut-t: [Redundant](https://github.com/povio/PovioKit/pull/356/commits/fd642483f4cb65cdbb467f8be60dbd8707d3a067) yes.

Sources/UI/SwiftUI/Views/RemoteImage/RemoteImage.swift
borut-t: [Agree](https://github.com/povio/PovioKit/pull/356/commits/74b6cfe17abdd2628924c91181edfd8130b6072e)

tonikocjan (approved):

PR #358, opened 2025-10-07
Chore/Split-out Networking

borut-t:
Since we are more and more using other networking clients, like openAPI or graphQL, bundling networking package together with its dependencies doesn't make sense. Therefore I've moved out PovioKitNetworking and PovioKitPromise into standalone [package](https://github.com/povio/PovioKitNetworking).


tonikocjan (approved):
Makes sense

PR #361, opened 2025-11-14
Fix/UserDefaults KVO

borut-t:
Made a change to handle scalar types without codable.

### DIFF
--- a/Sources/Utilities/PropertyWrapper/UserDefault.swift
+++ b/Sources/Utilities/PropertyWrapper/UserDefault.swift
@@ -122,14 +122,19 @@ private extension UserDefault {
   }
   
   func readPrimitive() -> Value? {
+    // For primitive types that never return nil (Bool, Int, Double, Float),
+    // we need to check if the key exists first, otherwise we'd return
+    // the UserDefaults default (false/0) instead of our defaultValue
+    let keyExists = storage.object(forKey: keyObject.key) != nil
+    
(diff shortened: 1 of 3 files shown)


Sources/Utilities/PropertyWrapper/UserDefault.swift
```diff
@@ -17,28 +17,39 @@ public struct UserDefault<Value: Codable> {
   
   public var wrappedValue: Value {
     get {
+      // try to read as primitive type first (for @AppStorage compatibility)
+      if let primitiveValue = readPrimitive() {
```
EgzonArifi, line 21: This primitive fast-path means Bool/Int/Double/Float will always return the value from UserDefaults (false/0 when the key is missing) and never fall back to defaultValue. For those types, defaultValue no longer represents “value when nothing is stored”, which is a breaking semantic change. We should probably restore a way to use defaultValue for the “missing key” case.

EgzonArifi (changes requested):

Sources/Utilities/PropertyWrapper/UserDefault.swift
borut-t: @EgzonArifi got it - https://github.com/povio/PovioKit/pull/361/commits/3bfb01cd646708d286098be9dbef1841675473cf

EgzonArifi (approved):
LGTM

